PR #93, opened 2016-01-11
Adjust spec for KeywordValue to be wrapper for string, add table of properties to StyleValue types.

wilddamon:
Apparently I actually did this last year and completely forgot!

Keywords are now just a wrapper for a String.

I also started a table for which properties can take which StyleValues, but got stuck with how to make a table in bikeshed. Any advice?

### DIFF
--- a/css-typed-om/Overview.bs
+++ b/css-typed-om/Overview.bs
@@ -39,16 +39,9 @@ Issue: is it possible to amalgamate this API with value objects once they're spe
 ============================================
 
 <pre class='idl'>
-enum StyleValueKeyword {"initial", "inherit", "revert", "unset"};
-
 interface StyleValue {
   attribute DOMString cssString;
   static (StyleValue or sequence&ltStyleValue>)? parse(DOMString property, DOMString cssText);
-};
-
-[Constructor(StyleValueKeyword)]
-interface KeywordValue : StyleValue {
-  attribute StyleValueKeyword keywordValue;
 };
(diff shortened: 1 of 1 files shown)


shans:
At a minimum, you can just use raw HTML. There might be a nicer way though - @tabatkins?

css-typed-om/Overview.bs
```diff
@@ -221,6 +211,19 @@ The value for a given property is the last valid value provided in the string.
...
+
+<pre class='idl'>
+[Constructor(DOMString)]
+interface KeywordValue : StyleValue {
+  attribute DOMString keywordValue;
+};
+</pre>
+
+{{KeywordValue}} objects represent {{StyleValue}}s that are keywords. It is simply a wrapper for a string
+and does not check that it contains a valid keyword.
```
shans: Add a note that property setters will enforce keyword validity.

tabatkins:
> but got stuck with how to make a table in bikeshed. Any advice?

Just write an HTML table. Bikeshed's source format is a subset of Markdown + HTML + some BS-specific stuff, so if it's not covered by vanilla Markdown, it's probably HTML.

Use good, semantic markup (thead/tbody, th, col/colgroup if appropriate) and then add `class=data` to get formatting automagically.  (fantasai put a lot of effort into the data-table styling and it shows.)

wilddamon:
Added some updates and a big partial table for accepted types.

PR #132, opened 2016-03-10
[css-typed-om] Add a typed OM spec for ColorValue

wilddamon:
@shans This is a tidied-up version of @njlawton's draft spec for ColorValue. PTAL


css-typed-om/Overview.bs
```diff
@@ -351,6 +351,29 @@ Issue: There are now 3 ways of creating Length values -
...
+---------------------------------------------
+<pre class='idl'>
+
+[Constructor(double r, double g, double b),
+ Constructor(double r, double g, double b, double a)]
+interface ColorValue : StyleValue{
+  readonly attribute double r;
+  readonly attribute double g;
+  readonly attribute double b;
+  readonly attribute double? a;
```
upsuper: Should `a` be nullable? I suppose it should just be `1.0` when not specified.

css-typed-om/Overview.bs
```diff
@@ -351,6 +351,29 @@ Issue: There are now 3 ways of creating Length values -
 
 Issue: What happens if the provided DOMString or LengthValue for (e.g.) a SimpleLength is invalid?
 
+{{ColorValue}} objects {#colorvalue-objects}
+---------------------------------------------
+<pre class='idl'>
+
+[Constructor(double r, double g, double b),
+ Constructor(double r, double g, double b, double a)]
```
upsuper: I wonder whether we should allow constructing `ColorValue` with supported color keywords. Or is there any existing function to do so?

wilddamon: @shans would have a better answer to that?

tabatkins:
Closing because we're instead porting over my previous attempt at adding a Color JS API from the Colors 4 spec.

PR #174, opened 2016-04-04
Add prose regarding SyntaxError

nainar:
#166

### DIFF
--- a/css-typed-om/Overview.bs
+++ b/css-typed-om/Overview.bs
@@ -294,13 +294,13 @@ dictionary CalcDictionary {
 };
 
 interface LengthValue : StyleValue {
-  LengthValue add(LengthValue value); // can throw
-  LengthValue subtract(LengthValue value); // can throw
-  LengthValue multiply(double value); // can throw
-  LengthValue divide(double value); // can throw
-  static LengthValue parse(DOMString cssText);
-  static LengthValue fromValue(double value, LengthType type);
-  static LengthValue fromDictionary(CalcDictionary dictionary);
+  LengthValue add(LengthValue value); // can throw SyntaxError
+  LengthValue subtract(LengthValue value); // can throw SyntaxError
+  LengthValue multiply(double value); // can throw SyntaxError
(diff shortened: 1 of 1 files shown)


css-typed-om/Overview.bs
```diff
@@ -336,6 +336,8 @@ interface SimpleLength : LengthValue {
 };
 </pre>
 
+If the provided DOMString or {{LengthValue}} is invalid the relevant function throws a SyntaxError.
```
shans: Need to describe why each method can throw.

e.g. add, etc. throw if called on or passed a LengthValue that isn't a SimpleLength or CalcLength.
parse throws if the DOMString isn't a valid representation of a length.
I don't think fromValue, fromDictionary can throw.

css-typed-om/Overview.bs
```diff
@@ -383,11 +383,11 @@ The following methods are defined for {{LengthValue}} objects:
 ::  Divides the length represented by the object by the provided <var>value</var>,
     and returns the result as a new {{LengthValue}}. This will construct
     a {{SimpleLength}} if the object is a {{SimpleLength}}, or a {{CalcLength}}
-    if the object is a {{CalcLength}}.
+    if the object is a {{CalcLength}}. The function can throw a DivideByZerError if value given is 0.
```
shans: "will throw a {{DivideByZeroError}} when the value given is 0."

css-typed-om/Overview.bs
```diff
@@ -383,11 +383,11 @@ The following methods are defined for {{LengthValue}} objects:
...
 
 :   <dfn method for=LengthValue>parse(DOMString cssText)</dfn>
 ::  Parses the provided <var>cssText</var> as a length value. Will return a
-    {{SimpleLength}} when possible, or a {{CalcLength}} otherwise.
+    {{SimpleLength}} when possible, or a {{CalcLength}} otherwise. The function can throw a SyntaxError if the DOMString it is passed doesn't represent a valid length.
```
shans: will throw, {{SyntaxError}}, when

Issue #146, opened 2016-04-04
[css-typed-om] Do we need to say that LengthValues must be positive?


shans:
e.g. for border-widths?

FremyCompany:
I think not. Setting a negative value for properties that do not accept them should just fail.

PR #175, opened 2016-04-04
Add note about LengthValues clamping at use time.

nainar:
#146

### DIFF
--- a/css-typed-om/Overview.bs
+++ b/css-typed-om/Overview.bs
@@ -356,6 +356,10 @@ interface SimpleLength : LengthValue {
 *   {{CalcLength}} objects represent lengths that contain multiple units (for example
     "calc(56em + 10%)").
 
+{{LengthValue}} objects are not range-restricted. Any valid combination of primitive lengths can be represented by a {{LengthValue}},
+that value will be accepted unaltered when set on a <a>specified StylePropertyMap</a> or
+<a>inline StylePropertyMap</a>. Instead, clamping and/or rounding will occur during computation of style.
+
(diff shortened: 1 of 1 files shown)


css-typed-om/Overview.bs
```diff
@@ -933,7 +933,9 @@ Shorthand properties and values are not supported.
 
 Issue(159): Spec up ColorValue
 
-Issue(146): Do we need to say that LengthValues must be positive e.g. for border-widths here?
+{{LengthValue}} objects are not range-restricted. Any valid number can be represented by a {{LengthValue}},
```
shans: 'Any valid combination of primitive lengths can be represented by a {{LengthValue}}'

css-typed-om/Overview.bs
```diff
@@ -933,7 +933,9 @@ Shorthand properties and values are not supported.
...
+and that value will not be clamped, rounded, or rejected when set on a <a>specified StylePropertyMap</a> or
```
shans: 'that value will be accepted unaltered when set on a'

css-typed-om/Overview.bs
```diff
@@ -933,7 +933,9 @@ Shorthand properties and values are not supported.
...
+<a>inline StylePropertyMap</a>. Instead, clamping and/or rounding will occur during computation of style.
```
shans: remove the 'Instead', part - replace with
'Properties which can't accept the full range of length values (e.g. 'border-width') perform clamping and/or rounding during style computation.'.

PR #214, opened 2016-06-07
s/scratch bitmap/output bitmap/ to fix broken link

dholbert:
The spec uses the term "scratch bitmap", which links to
 https://html.spec.whatwg.org/multipage/scripting.html#scratch-bitmap
which is a broken link.

I believe this is meant to point to the definition of "output bitmap" in that HTML spec, i.e. it should be pointing to:
 https://html.spec.whatwg.org/multipage/scripting.html#output-bitmap
...and it should be using the consistent term (with the word "output" instead of "scratch").

Not sure if the HTML spec changed out from under us or what, but anyway, this should get us back to a state with working links.


bfgeek:
Yeah HTML spec changed. Thanks ^-^

PR #251, opened 2016-06-24
[worklets] First pass of fixing import for worklets.

bfgeek:
@domenic A first pass of fixing worklets import using a shared cache. Wording for the cache is very similar to the module map in html, and seems ok?

Not happy with "fetched module map" naming.

cc/ @bzbarsky @annevk

### DIFF
--- a/worklets/Overview.bs
+++ b/worklets/Overview.bs
@@ -13,6 +13,9 @@ Editor: Ian Kilpatrick, ikilpatrick@chromium.org
 <pre class="anchors">
 urlPrefix: http://heycam.github.io/webidl/#dfn-; type: dfn;
     text: inherit
+urlPrefix: https://fetch.spec.whatwg.org/; type: dfn;
+    urlPrefix: #concept-;
+        text: fetch
 urlPrefix: https://html.spec.whatwg.org/multipage/browsers.html; type: dfn;
     text: effective script origin
     url: #origin-2; text: origin
@@ -247,7 +250,7 @@ The {{Worklet}} object provides the capability to import module scripts into its
 
 <pre class='idl'>
 interface Worklet {
-    [NewObject] Promise&lt;void> import(DOMString moduleURL);
+    [NewObject] Promise&lt;void> import(USVString moduleURL);
 };
 </pre>
 
@@ -262,39 +265,72 @@ is empty; it is populated when the user agent chooses to create its {{WorkletGlo
 A {{Worklet}} has a list of the <dfn>worklet's resolved module URLs</dfn>. Initially this list is
 empty; it is populated when module scripts resolved.
 
+A {{Worklet}} has a <dfn>fetched modules map</dfn>. This is a map of <a>fetch</a> requests to values
+that are a <a>fetch</a> responses. Access to this map should be thread-safe.
+
+The <a>fetched modules map</a> exists to ensure that {{WorkletGlobalScope}}s created at different
+times contain the same set of scripts and have the same behaviour. The creation of additional
+{{WorkletGlobalScope}}s should be transparent to the author.
+
 When the <dfn method for=Worklet>import(|moduleURL|)</dfn> method is called on a {{Worklet}} object,
 the user agent <em>must</em> run the following steps:
     1. Let |promise| be <a>a new promise</a>.
 
-    2. Run the following steps <a>in parallel</a>:
-        1. Let |resolvedModuleURL| be the result of <a>resolving</a> the |moduleURL| relative to the
-            <a>API base URL</a> specified by the <a>entry settings object</a> when the method was
-            invoked.
+    2. Let |resolvedModuleURL| be the result of <a>resolving</a> the |moduleURL| relative to the
+        <a>API base URL</a> specified by the <a>entry settings object</a> when the method was
+        invoked.
 
-        2. If this fails, reject |promise| with a <a>SyntaxError</a> exception and abort these
-            steps.
+    3. If this fails, reject |promise| with a <a>SyntaxError</a> exception and abort these steps.
 
-        3. Add |resolvedModuleURL| to the list of <a>worklet's resolved module URLs</a>.
+    4. Ensure that there is at least one {{WorkletGlobalScope}} in the <a>worklet's
+        WorkletGlobalScopes</a>. If not <a>create a WorkletGlobalScope</a> given the current
+        {{Worklet}}.
 
-        4. Ensure that there is at least one {{WorkletGlobalScope}} in the <a>worklet's
-            WorkletGlobalScopes</a>. If not <a>create a WorkletGlobalScope</a> given the current
-            {{Worklet}}.
+        The user-agent may also create additional {{WorkletGlobalScope}}s at this time.
 
-        5. For each {{WorkletGlobalScope}} in the <a>worklet's WorkletGlobalScopes</a>, run these
+    5. Add |resolvedModuleURL| to the list of <a>worklet's resolved module URLs</a>.
+
+    6. Let |outsideSettings| be the <a>incumbent settings object</a>.
+
+    7. Run the following steps <a>in parallel</a>:
+
+        1. For each {{WorkletGlobalScope}} in the <a>worklet's WorkletGlobalScopes</a>, run these
             substeps:
-            1. Let |settings| be the {{WorkletGlobalScope}}'s associated <a>environment settings
-                object</a>.
 
-            2. Let |script| be the result of <a>fetch a module script tree</a> given
-                |resolvedModuleURL|, "anonymous" for the <a>CORS setting attribute</a>, and
-                |settings|.
+            1. Let |insideSettings| be the {{WorkletGlobalScope}}'s associated <a>environment
+                settings object</a>.
 
-            Note: Worklets follow <a>web workers</a> here in not allowing "use-credientials" for
-                fetching resources.
+            2. <a>Fetch a module script tree</a> given |resolvedModuleURL|, "omit", the empty string
+                (as no cryptographic nonce is present for worklets), "not parser-inserted",
+                "script", |outsideSettings|, and |insideSettings|.
 
-            3. <a>Run a module script</a> given |script|.
+                - custom fetch hook, given |request|:
 
-        6. If <em>all</em> the steps above succeeded (in particular, if all of the scripts parsed
+                    1. Let |cache| be the current {{Worklet}}'s <a>fetched modules map</a>.
+
+                    2. If |cache| contains an entry with key |request| whose value is "fetching",
+                        wait (<a>in parallel</a>) until that entry's value changes, then proceed to
+                        the next step.
+
+                    3. If |cache| contains an entry with key |request|, asynchronously complete this
+                        algorithm with that entry's value, and abort these steps.
+
+                    4. Create an entry in |cache| with key |request| and value "fetching".
+
+                    5. <a>Fetch</a> |request|.
+
+                    6. Let |response| be the result of <a>fetch</a> when it asynchronously
+                        completes.
+
+                    7. Set the value of the entry in |cache| whose key is |request| to |response|,
+                        and asynchronously complete this algorithm with |response|.
+
+            3. Let |script| be the result of <a>fetch a module script tree</a> when it
+                asynchronously completes.
+
+            4. <a>Run a module script</a> given |script|.
+
+        2. If <em>all</em> the steps above succeeded (in particular, if all of the scripts parsed
                 and loaded into the global scopes), resolve |promise|.
                 <br>Otherwise, reject |promise|.
 
@@ -302,9 +338,10 @@ the user agent <em>must</em> run the following steps:
             reject the promise; if the script throws an error while first evaluating the promise
             should resolve as a classes may have been registered correctly.
 
-    3. Return |promise|.
+    8. Return |promise|.
 
-Issue(w3c/css-houdini-drafts#47): Need ability to load code into {{WorkletGlobalScope}} declaratively.
+Issue(w3c/css-houdini-drafts#47): Need ability to load code into a {{WorkletGlobalScope}}
+    declaratively.
 
 Lifetime of the Worklet {#lifetime-of-the-worklet}
 --------------------------------------------------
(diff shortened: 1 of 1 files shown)


worklets/Overview.bs
```diff
@@ -262,49 +265,83 @@ is empty; it is populated when the user agent chooses to create its {{WorkletGlo
 A {{Worklet}} has a list of the <dfn>worklet's resolved module URLs</dfn>. Initially this list is
 empty; it is populated when module scripts resolved.
 
+A {{Worklet}} has a <dfn>fetched modules map</dfn>. This is a map of <a>fetch</a> requests to values
```
domenic: Maybe "module responses map"

worklets/Overview.bs
```diff
@@ -262,49 +265,83 @@ is empty; it is populated when the user agent chooses to create its {{WorkletGlo
...
+that are a <a>fetch</a> responses. Access to this map should be thread-safe.
+
+The <a>fetched modules map</a> exists to ensure that {{WorkletGlobalScope}}s created at different
+times contain the same set of scripts and have the same behaviour. The creation of additional
```
domenic: "same set of scripts" is a bit ambiguous, since at least in the spec sense, they have different scripts. It's more about using the same source text, and not doing double-fetches, I believe.

worklets/Overview.bs
```diff
@@ -262,49 +265,83 @@ is empty; it is populated when the user agent chooses to create its {{WorkletGlo
...
 When the <dfn method for=Worklet>import(|moduleURL|)</dfn> method is called on a {{Worklet}} object,
 the user agent <em>must</em> run the following steps:
     1. Let |promise| be <a>a new promise</a>.
 
-    2. Run the following steps <a>in parallel</a>:
-        1. Let |resolvedModuleURL| be the result of <a>resolving</a> the |moduleURL| relative to the
-            <a>API base URL</a> specified by the <a>entry settings object</a> when the method was
-            invoked.
+    2. Let |resolvedModuleURL| be the result of <a>resolving</a> the |moduleURL| relative to the
+        <a>API base URL</a> specified by the <a>entry settings object</a> when the method was
```
domenic: [Don't use entry settings object.](https://readable-email.org/list/public-script-coord/topic/multiple-globals-and-you) Use current settings object. Probably the same, in a worklet...

worklets/Overview.bs
```diff
@@ -262,49 +265,83 @@ is empty; it is populated when the user agent chooses to create its {{WorkletGlo
...
+        invoked.
+
+    3. If this fails, reject |promise| with a <a>SyntaxError</a> exception and abort these steps.
```
domenic: Probably DOMException "SyntaxError", not just SyntaxError?

worklets/Overview.bs
domenic: You want this to be a map of request URLs -> responses. Because "fetch a module script tree" will give you a new request every time, so you can't use the requests as cache keys. This also allows you to eliminate the "worklet's list of resolved module URLs"

worklets/Overview.bs
```diff
@@ -262,49 +265,83 @@ is empty; it is populated when the user agent chooses to create its {{WorkletGlo
...
-            steps.
+        The user-agent may also create additional {{WorkletGlobalScope}}s at this time.
 
-        3. Add |resolvedModuleURL| to the list of <a>worklet's resolved module URLs</a>.
+    5. Add |resolvedModuleURL| to the list of <a>worklet's resolved module URLs</a>.
 
-        4. Ensure that there is at least one {{WorkletGlobalScope}} in the <a>worklet's
-            WorkletGlobalScopes</a>. If not <a>create a WorkletGlobalScope</a> given the current
-            {{Worklet}}.
+    6. Let |outsideSettings| be the <a>incumbent settings object</a>.
```
domenic: Use current settings object instead.

worklets/Overview.bs
```diff
@@ -262,49 +265,83 @@ is empty; it is populated when the user agent chooses to create its {{WorkletGlo
...
+
+            2. <a>Fetch a module script tree</a> given |resolvedModuleURL|, "omit", the empty string
+                (as no cryptographic nonce is present for worklets), "not parser-inserted",
+                "script", |outsideSettings|, and |insideSettings|.
+
+                - custom fetch hook, given |request|:
+
+                    1. Let |cache| be the current {{Worklet}}'s <a>fetched modules map</a>.
+
+                    2. If |cache| contains an entry with key |request| whose value is "fetching",
```
domenic: Use request's url as the key

domenic:
Core of the algorithm looks pretty good. Some minor suggestions leading up to it. I'll get working on that patch to HTML now.

worklets/Overview.bs
bzbarsky: This call isn't happening in a worklet, so this is not the same at all.

Current settings isn't used by _anything_ I'm aware of as a source of base URLs in the web platform.  Some things use entry and some use relevant settings of `this`.  Let's not add a third way to do base URLs, please.

domenic: Happy to use relevant settings of `this`. Thanks for jumping in.

worklets/Overview.bs
bzbarsky: Again, why not relevant settings?  What is this actually being used for?

worklets/Overview.bs
bfgeek: Changed to "module responses map". I've still got "worklet's list of resolved module URLs" at the moment as the load ordering when initializing the worklet is still important.

Could change this to just an "ordered map"?

worklets/Overview.bs
bfgeek: Yup, updated to "same set of script source text".

worklets/Overview.bs
bfgeek: PTAL

worklets/Overview.bs

worklets/Overview.bs

worklets/Overview.bs
domenic: Oh, yeah, that sounds good to me. In JS all maps are ordered so that was natural for me :). Maybe say things like "the list of keys of the module responses map, in insertion order".

worklets/Overview.bs
```diff
@@ -259,52 +269,84 @@ Note: As an example the <a>worklet global scope type</a> might be a {{PaintWorkl
...
 When the <dfn method for=Worklet>import(|moduleURL|)</dfn> method is called on a {{Worklet}} object,
 the user agent <em>must</em> run the following steps:
     1. Let |promise| be <a>a new promise</a>.
 
-    2. Run the following steps <a>in parallel</a>:
-        1. Let |resolvedModuleURL| be the result of <a>resolving</a> the |moduleURL| relative to the
-            <a>API base URL</a> specified by the <a>entry settings object</a> when the method was
-            invoked.
+    2. Let |resolvedModuleURL| be the result of <a>parsing</a> the |moduleURL| argument relative to
+        the <a>settings object</a> of <b>this</b>.
```
domenic: You want [relevant settings object](https://html.spec.whatwg.org/multipage/webappapis.html#relevant-settings-object) of `this`.

domenic:
@bzbarsky let's talk about settings objects. I'm starting from the assumption we want to stick to current or relevant.

The code in question is:

```js
window1.Worklet.prototype.import.call(window2.someWorklet, url);
```

**URL resolution**: Current means this resolves relative to `window1`'s document's URL, relevant means `window2`. I don't have a strong preference, but to me current seems like a better fit for our general thesis of "current by default, relevant if creating long-lived objects that are stored somewhere." We're not creating any objects here, so to me current makes more sense. And that holds true in general for URL resolution; I'd hope to update the rest of the specs in that direction too.

**Picking the "outside settings object"**: The question is whether the fetch client for the module requests is `window1` or `window2`. Again I feel like our general rule guides us toward current. This is complicated by the fact that @bfgeek's previous use of incumbent was trying to be consistent with [the Worker constructor](https://html.spec.whatwg.org/multipage/workers.html#dom-worker), which uses the incumbent settings object as the fetch client. But we haven't tested browsers and I'm hoping that can change---again, I'd assume ideally to current.

worklets/Overview.bs

worklets/Overview.bs

domenic:
> Picking the "outside settings object"

I ran a test and it appears that the Worker constructor's choice of incumbent does not match browsers, which all use current. (At least, Firefox and Chrome do; Edge doesn't seem to send referrer headers so I can't test, and I haven't tested Safari yet.) See https://settings-object-worker-client-ckrcsrtdqr.now.sh/entry.html for a test.

So that further implies current is correct for picking the outside settings object. I'll update Workers tomorrow in HTML.

I'll also test URL resolution in the Worker constructor tomorrow.

bzbarsky:
There's a fundamental difference between constructors and instance methods: with constructors, the current settings object _is_ the relevant settings object of the thing being constructed.  With instance methods, the two are not necessarily related.

I think our general rule as stated is probably wrong.  It's the general rule in ES for lack of anything else because ES doesn't have a concept of "relevant", but in practice ES has all sorts of hooks for overriding the behavior for instance methods (using Symbol.species and the like).  The current global mostly gets used in observable ways in ES as a fallback when other, saner, things have failed.

In this particular case, it seems to me that relative URL resolution for a worklet would naturally be a property of the worklet itself, not of the function you happened to call to trigger the load.

Or put another way, if you import two things into the same worklet but using `import` methods from different globals for whatever reason (probably because your libraries are doing something bizarre), would you expect different base URLs to be used or the same base URL?  For XHR, the same base URL is used for `open` in browsers if you call `open` multiple times, because XHR uses the relevant settings of `this` (the thing actually doing the load) for the base URL...

domenic:
(Sorry for cluttering up your pull request, @bfgeek!)

> I think our general rule as stated is probably wrong. 

Hmm. I don't really agree, because I think the idea of objects having realms is a bit unnatural. As a developer I think if anything I'd expect the object's realm to that detected via its `__proto__` chain (i.e., which `Object.prototype` does it bottom out in). But that's not what relevant realm is; relevant realm is immutable over time. In contrast, it makes sense to me to think of "what is the realm in which the code that's running was 'declared'", i.e. it makes sense for functions to have realms.

In other words, putting on my naive developer hat, I'd conceptualize `import` to be written as something like

```js
Worklet.prototype.import = function (url) {
  const resolvedURL = new URL(url, location.href).href;  // (1)
  // ...
  fetch(resolvedURL, ...).then(...); // (2)
}
```

Here at both (1) and (2) I am implicitly using "current". I'd have no way to get "relevant" from JS code; it's fairly unnatural to think of doing so.

> In this particular case, it seems to me that relative URL resolution for a worklet would naturally be a property of the worklet itself, not of the function you happened to call to trigger the load.

In contrast, to me it seems that the function that I happened to call to trigger the load is the one doing URL resolution, so it should be a property of the function, since that's what's actually executing the load.

> Or put another way, if you import two things into the same worklet but using import methods from different globals for whatever reason (probably because your libraries are doing something bizarre), would you expect different base URLs to be used or the same base URL?

When you make it concrete like this, I can't say I have any particular expectations. Both seem equally weird to me. You'd be forcing me as a developer to confront the question of "what does a relative URL really mean in a multi-global context," and I don't have a good intuitive answer. I can only fall back to more principled arguments like the above.

---

That said, I'm OK changing our position here. If you find my above arguments unpersuasive, then I'd like to get the resulting advice written up into HTML's multi-globals section. I can't tell if you're advocating for a more general "relevant by default, current when nothing else makes sense" position, or specifically for saying that URL resolution should use a relevant object.

(We should probably also include a note that in constructors, relevant = current.)

bzbarsky:
> I don't really agree, because I think the idea of objects having realms is a bit unnatural.

I honestly don't see why.  Especially for IDL objects, which are explicitly specified as "having realms".

But more to the point, it's not whether objects "have realms" or not but whether they are associated with some particular state that may happen to be associated with a Realm, or may not be.  In this case, a base URL, or a source of such.

> I think if anything I'd expect the object's realm to be that detected via its __proto__ chain

That's obviously not workable in general because the proto chain may not lead to anywhere.... And yes, I know that some ES builtin objects work if you null out their proto and some start misbehaving badly, basically depending on whether people tried to make things work that way or not.  It's quite inconsistent and a bit annoying, really.

> In other words, putting on my naive developer hat, I'd conceptualize import to be written as something like

That's fair, except that's not how other things that do base URLs in the platform work.  Ignoring for the moment that `location.href` has nothing to do with the base URL of anything in general, they effectively (in the naive developer sense) capture the thing that has the base URL they care about in a closure when they are created and use it thereafter.

> I'd have no way to get "relevant" from JS code

You can think of "relevant" as a symbol-named property on the `this`, or something that symbol-named "non-public" functions that do the real work close over.

Or heck, a value stored in a weakmap.

> it seems that the function that I happened to call to trigger the load is the one doing URL resolution

Again, that's not how anything works in the web platform right now.  In practice the model is that all the API functions call some sort of internal functions that do the actual work.  This is in some sense an accident of history (in that the internal functions are implemented in C++ and whatnot), and we can obviously have an argument about how important consistency is.  But that's how things work right now, and trying to change the model piecemeal is not very appealing to me, precisely because it means we throw consistency out the window forever.

> I can't tell if you're advocating for a more general "relevant by default, current when nothing else makes sense" position

I'm not sure whether I am.  I'd have to see what things do in practice, and in particular what uses people actually have for settings objects.  We have some obvious uses: base URLs (uses entry or relevant), initial prototypes for newly created objects (largely uses relevant in browsers, but uses current in some cases when the creation is done by the binding layer itself; think rejected promises when the `this` is the wrong type).  Fetch clients, I have no idea what browsers do.  Other uses?

It seems to me that such a survey is well worth doing before deciding what our general advice should be, because we should be valuing compatibility and consistency over theoretical purity...

> (We should probably also include a note that in constructors, relevant = current.)

As usual, "sort of".  The relevant settings of the object that gets created == current.  But the relevant settings of `new.target` of course might not match current.  Whether anyone cares about the relevant settings of `new.target` in a constructor ... I don't know.

Note that I'm on vacation for the next week, so responses will be a bit laggy, sorry.

domenic:
So I think we should probably go with relevant here since @bzbarsky seems to strongly prefer that. I don't want to block this getting merged.

bfgeek:
Thanks for the review all! Changed outsideSettings to relevant settings object of this.

PR #253, opened 2016-06-29
Fix extremely important article use before 'box'

andymutton:
I am pedantic and I can't handle it.


tabatkins:
Pedantry is the best reason for a PR.

PR #264, opened 2016-07-20
[worklets] Fix module loading when creating a new worklet global scope.

bfgeek:
@domenic can you take a quick look at this?

### DIFF
--- a/worklets/Overview.bs
+++ b/worklets/Overview.bs
@@ -18,6 +18,7 @@ urlPrefix: http://heycam.github.io/webidl/#dfn-; type: dfn;
 urlPrefix: https://fetch.spec.whatwg.org/; type: dfn;
     urlPrefix: #concept-;
         text: fetch
+        url: request-url; text: url;
 urlPrefix: https://html.spec.whatwg.org/multipage/browsers.html; type: dfn;
     text: effective script origin
     url: #origin-2; text: origin
@@ -180,18 +181,16 @@ When a user agent is to <dfn>create a WorkletGlobalScope</dfn>, for a given |wor
 
     4. Associate the |settingsObject| with |workletGlobalScope|.
 
-    5. For each |entry| in the given |worklet|'s <a>module responses map</a> (in insertion order),
-        run the following substeps:
+    5. For each |resolvedModuleURL| in the given |worklet|'s <a>top level module url list</a> (in
+        insertion order), run the following substeps:
 
-        1. Let |resolvedModuleURL| be |entry|'s key.
+        1. <a>Fetch a module worklet script tree</a> given |worklet|, |resolvedModuleURL|,
+            |outsideSettings|, and |settingsObject|.
 
-        2. <a>Fetch a module worklet script tree</a> given |resolvedModuleURL|, |outsideSettings|,
-            and |settingsObject|.
-
-        3. Let |script| be the result of <a>fetch a module worklet script tree</a> when it
+        2. Let |script| be the result of <a>fetch a module worklet script tree</a> when it
             asynchronously completes.
 
-        4. <a>Run a module script</a> given |script|.
+        3. <a>Run a module script</a> given |script|.
 
 ### Script settings for worklets ### {#script-settings-for-worklets}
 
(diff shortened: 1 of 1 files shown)


worklets/Overview.bs
```diff
@@ -185,14 +185,13 @@ When a user agent is to <dfn>create a WorkletGlobalScope</dfn>, for a given |wor
 
         1. Let |resolvedModuleURL| be |entry|'s key.
 
-        2. Let |script| be the result of <a>fetch a module script tree</a> given
-            |resolvedModuleURL|, "anonymous" for the <a>CORS setting attribute</a>, and
-            |settingsObject|.
+        2. <a>Fetch a module worklet script tree</a> given |resolvedModuleURL|, |outsideSettings|,
```
bfgeek: Slight problem as we don't have access to |outsideSettings| here.

worklets/Overview.bs
```diff
@@ -247,6 +246,41 @@ When a user agent is to <dfn>set up a worklet environment settings object</dfn>,
 
 Issue: Merge this with https://html.spec.whatwg.org/multipage/workers.html#set-up-a-worker-environment-settings-object
 
+### Fetching a module worklet script tree ### {#fetching-a-module-worklet-script-tree}
+
+When a user agent is to <dfn>fetch a module worklet script tree</dfn>, given |resolvedModuleURL|,
+|outsideSettings|, and |insideSettings| it must run the following steps:
```
domenic: Nit: comma after `|insideSettings|`

worklets/Overview.bs
```diff
@@ -247,6 +246,41 @@ When a user agent is to <dfn>set up a worklet environment settings object</dfn>,
...
+
+    1. <a>Fetch a module script tree</a> given |resolvedModuleURL|, |outsideSettings|, "script" the
+        empty string (as no cryptographic nonce is present for worklets), "not parser-inserted", "omit",
+        and |insideSettings|.
```
domenic: So, things have changed here a bit. I think you want [fetch a module worker script tree](http://localhost:8080/#fetch-a-module-worker-script-tree) given resolvedModuleURL, outsideSettings, "script", "omit", and insideSettings.

Also, you may want to invent a new destination of "worklet", since there are separate "sharedworker" and "worker" destinations, so it seems like that's the level of granularity we're operating at. That would require a change to Fetch (@annevk).

worklets/Overview.bs
```diff
@@ -247,6 +246,41 @@ When a user agent is to <dfn>set up a worklet environment settings object</dfn>,
...
+
+        To <a>perform the fetch</a> given |request|, perform the following steps:
+
+            1. Let |cache| be the current {{Worklet}}'s <a>module responses map</a>.
```
domenic: "current Worklet" doesn't seem very defined. Probably insideSetting's [global object](http://localhost:8080/#concept-settings-object-global).

worklets/Overview.bs
```diff
@@ -247,6 +246,41 @@ When a user agent is to <dfn>set up a worklet environment settings object</dfn>,
...
+
+            2. Let |url| be |request|'s <a >url</a>.
```
domenic: Nit: extra space in `<a >`

domenic: Also, this goes to the wrong place; you want to go to https://fetch.spec.whatwg.org/#concept-request-url

worklets/Overview.bs
domenic: I don't understand this whole section. The module responses map stores every module, including imported ones. And you want to fetch and execute them all? That means that if you do `import './a.js'` and a.js does `import './b.js'`, then the module responses map will have both a.js and b.js in it. Then you will execute the tree starting at a.js (so both a.js and b.js), but also you will separately execute the tree starting at b.js (so b.js will be executed twice). This seems bad?

What is the intention of this section?

worklets/Overview.bs

worklets/Overview.bs
bfgeek: Changed. I was discussing this with @mikewest (as destination is just used for the CSP check right?) settled on "script-src" being ok initially.

Let me know if I'm off the track here in what this param does. (traced it through the other day... i think)

worklets/Overview.bs
bfgeek: The cache lives of the Worklet object, not the global, but done by passing other parameter into this function.

worklets/Overview.bs
bfgeek: Done & done.

worklets/Overview.bs
bfgeek: Yeah this was wrong. This only should be loading the top level requests.

This section is used for if a user-agent wants to create a new WorkletGlobalScope in addition to the ones it already has. (E.g. it may have killed one from the pool and needs to restart or something).

When instantiating the new WorkletGlobalScope it has to load all the previous top level modules loaded into it.

Added a "top level module url list" to the worklet which the fetch hook now populates. This reads off that to load it's code.

worklets/Overview.bs
domenic: Hmm. "script-src" is the CSP directive currently used for `<script>` and `importScripts()`, whereas workers and shared workers use "child-src". The choice here is also web-exposed when a service worker intercepts the request. Maybe open a follow-up issue so we can get everyone on the same page?

worklets/Overview.bs
```diff
@@ -336,8 +342,8 @@ the user agent <em>must</em> run the following steps:
             1. Let |insideSettings| be the {{WorkletGlobalScope}}'s associated <a>environment
                 settings object</a>.
 
-            2. <a>Fetch a module worklet script tree</a> given |resolvedModuleURL|,
-                |outsideSettings|, and |settingsObject|.
+            2. <a>Fetch a module worklet script tree</a> given the current {{Worklet}},
```
domenic: Nit: "the current Worklet" is not defined here. I think you want to add a step at the top of this method definition that does "Let _worklet_ be this `Worklet` object" or similar. (Other acceptable phrasing: "the `Worklet` object on which this method was called".) Then you can use _worklet_ later.

domenic:
LGTM with one minor thing and the open question about destination. Awesome!

domenic:
With https://github.com/whatwg/html/pull/6056, we can close this. That PR incorporates a revised version of this patch; notably, this patch was missing any step to actually add URLs to the top-level URL list, but I added that in to HTML.

PR #271, opened 2016-07-28
[css-typed-om] Make CSSSkew and CSSRotation store their angle members as CSSAngleValues instead of doubles

rjwright:
This makes them more consistent with CSSPerspective and CSSTranslation, which store their length members and CSSLengthValues.


rjwright:
@shans

Issue #192, opened 2016-05-12
[css-typed-om] CSSTokenStream values need to computed to other values in some cases


tabatkins:
Spec currently says that compute to another CSSTokenStream, just with all the variables resolved. This is true for custom properties, but when any other property is a CSSTokenStream, the computed value is whatever types that property normally has.

shans:
It's not even true for custom properties that have a registered type.

shans:
I think the spec actually doesn't say this at all. I rewrote the section to be a lot clearer though.

shans:
@tabatkins ptal at attached pull request

shans:
Closed via #296

PR #296, opened 2016-09-12
[css-typed-om] clarify how CSSUnparsedValues compute.

shans:
(https://github.com/w3c/css-houdini-drafts/issues/192)

### DIFF
--- a/css-typed-om/Overview.bs
+++ b/css-typed-om/Overview.bs
@@ -1419,28 +1419,40 @@ appear as computed values (i.e. as a value stored on computed
 Computed {{CSSUnparsedValue}} objects {#computed-unparsedvalue-objects}
 -----------------------------------------------------------------------------
 
-Custom property references are resolved as part of style computation. Accordingly,
-computed {{CSSUnparsedValue}} objects will not contain {{CSSVariableReferenceValue}} objects.
-As a result, only a single {{DOMString}} will appear in the sequence contained by
-computed {{CSSUnparsedValue}} objects.
+A property with a specified {{CSSUnparsedValue}} value will not
+compute to a {{CSSUnparsedValue}}. Instead, after custom property references
+are resolved, the {{CSSStyleValue}} subclass appropriate to the property will be
+used.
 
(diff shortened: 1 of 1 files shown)


css-typed-om/Overview.bs
```diff
@@ -1419,28 +1419,40 @@ appear as computed values (i.e. as a value stored on computed
...
 
-Custom property references are resolved as part of style computation. Accordingly,
-computed {{CSSUnparsedValue}} objects will not contain {{CSSVariableReferenceValue}} objects.
-As a result, only a single {{DOMString}} will appear in the sequence contained by
-computed {{CSSUnparsedValue}} objects.
-
-Furthermore, values that at specified value time contained custom property references
-are renormalized after computation.
+A property with a specified {{CSSUnparsedValue}} value will not
+compute to a {{CSSUnparsedValue}}. Instead, after custom property references
```
tabatkins: will not automatically compute

Because they *can* still compute to that, if they're custom properties without a syntax.

css-typed-om/Overview.bs
```diff
@@ -1419,28 +1419,40 @@ appear as computed values (i.e. as a value stored on computed
...
+computed width will be represented by a {{CSSLengthValue}} object (assuming
+that --foo resolves to a valid substitution).
+</div>
 
-<pre class='lang-javascript'>
-var a = e.styleMap.get('width');
-var b = getComputedStyleMap(e).get('width');
-</pre>
+Often there will be no {{CSSStyleValue}} subclass appropriate - for example when a custom property
+contains a reference to another custom property. In these cases, a {{CSSStyleValue}}
```
tabatkins: CSSUnparsedValue

tabatkins: Just, like, a plain CSSStyleValue? Why not leave it as CSSUnparsedValue?

shans:
Discussed with @tabatkins offline - decided that it's correct to use CSSUnparsedValue only when there are unresolved variable references and CSSStyleValue otherwise.

PR #336, opened 2017-01-04
[css-layout-api] Fix typo where the wrong variable was accessed.

emilio:
Sorry for the mini-patch, I just spotted that and thought it was worth to fix.


tabatkins:
Mini-patches are very much appreciated! Thanks!

PR #337, opened 2017-01-05
[css-paint-api] Adds paint function arguments.

### LINKED ISSUE #100: [css-paint-api] Allow paint function arguments, properties, or both?
Example with just properties:

```
.foo {
  --bar: 5;
  background-image: paint(foo), paint(foo);
}

registerPaint(class {
  static inputProperties = ['--bar', 'color'];
  void paint(ctx, geom, styleMap) {
    // --bar & color both available here.
  }
});
```

Pros:
- Can use `color`.

Cons:
- Can't use same function to get different result (like `linear-gradient`).

Example with just input arguments

```
.foo {
  --bar: 5;
  background-image: paint(foo, var(--bar)), paint(foo, 10);
}

registerPaint(class {
  static inputArguments = ['&lt;number&gt;'];
  void paint(ctx, geom, styleMap) {
    // --bar & color both available here.
  }
});
```

Pros:
- Can use same function with different arguments.

Cons:
- Can't access native properties like `color`.

bfgeek:
E.g.

registerPaint('foo', class {
    static get inputArguments() { return ['<length>']; }

    paint(ctx, size, styleMap, arguments) {

    }
});

Fixes #100.

### DIFF
--- a/css-paint-api/Overview.bs
+++ b/css-paint-api/Overview.bs
@@ -14,6 +14,20 @@ Editor: Dean Jackson, dino@apple.com
 
 <pre class="link-defaults">
 spec:css-break-3; type:dfn; text:fragment
+spec:css-ui-3; type:property; text:cursor
+</pre>
+
+<pre class='biblio'>
+{
+    "css-properties-values-api": {
+        "title": "CSS Properties and Values API",
+        "href": "https://drafts.css-houdini.org/css-properties-values-api/"
+    },
+    "css-typed-om": {
+        "title": "CSS Typed OM",
+        "href": "https://drafts.css-houdini.org/css-typed-om/"
+    }
+}
 </pre>
 
 <pre class="anchors">
@@ -38,6 +52,7 @@ urlPrefix: https://tc39.github.io/ecma262/#sec-; type: dfn;
     text: IsConstructor
     text: HasProperty
     url: get-o-p; text: Get
+    url: ecmascript-data-types-and-values; text: Type
     url: terms-and-definitions-function; text: function
     urlPrefix: native-error-types-used-in-this-standard-
         text: TypeError
@@ -60,6 +75,9 @@ Paint Invalidation {#paint-invalidation}
 
 A <a>document</a> has an associated <dfn>paint name to input properties map</dfn>. Initially it is
 empty and is populated when {{registerPaint(name, paintCtor)}} is called.
+
+A <a>document</a> has an associated <dfn>paint name to input arguments syntax map</dfn>. Initially
+it is empty and is populated when {{registerPaint(name, paintCtor)}} is called.
 
 Each <<paint()>> function for a box has an associated <dfn>paint valid flag</dfn>. It may be either
 <dfn>paint-valid</dfn> or <dfn>paint-invalid</dfn>. It is initially set to <a>paint-invalid</a>.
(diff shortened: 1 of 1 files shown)


bfgeek:
@tabatkins can you check the language here for input arguments?

css-paint-api/Overview.bs
```diff
@@ -180,32 +199,52 @@ called, the user agent <em>must</em> run the following steps:
...
 
-    9. Let |prototype| be the result of <a>Get</a>(|paintCtor|, "prototype").
+    14. Let |prototype| be the result of <a>Get</a>(|paintCtor|, "prototype").
 
-    10. If the result of <a>Type</a>(|prototype|) is not Object, <a>throw</a> a <a>TypeError</a> and
+    15. If the result of <a>Type</a>(|prototype|) is not Object, <a>throw</a> a <a>TypeError</a> and
         abort all these steps.
 
-    11. Let |paint| be the result of <a>Get</a>(|prototype|, "paint").
+    16. Let |paint| be the result of <a>Get</a>(|prototype|, "paint").
```
tabatkins: Does this mean that it's *required* that "paint" be on the prototype, rather than on a higher prototype? Is there a more abstract operation we can use instead that just walks the proto chain? If so, we should just use that - there's no reason it shouldn't be able to show up on a higher proto, or on the instance.

css-paint-api/Overview.bs
```diff
@@ -180,32 +199,52 @@ called, the user agent <em>must</em> run the following steps:
...
 
-    12. If the result of <a>IsCallable</a>(|paint|) is false, <a>throw</a> a <a>TypeError</a> and
+    17. If the result of <a>IsCallable</a>(|paint|) is false, <a>throw</a> a <a>TypeError</a> and
```
tabatkins: I *think* this can be replaced with use of the IDL callback coercion rules <https://heycam.github.io/webidl/#es-callback-function>

css-paint-api/Overview.bs
```diff
@@ -61,6 +76,9 @@ Paint Invalidation {#paint-invalidation}
 A <a>document</a> has an associated <dfn>paint name to input properties map</dfn>. Initially it is
 empty and is populated when {{registerPaint(name, paintCtor)}} is called.
 
+A <a>document</a> has an associated <dfn>paint name to input arguments syntax map</dfn>. Initially
```
tabatkins: Spec that it's an Infra Standard map? <https://infra.spec.whatwg.org/#maps>

css-paint-api/Overview.bs
```diff
@@ -216,11 +255,14 @@ called, the user agent <em>must</em> run the following steps:
         - <a>paint constructor valid flag</a> being true
 
         - <a>paint context alpha flag</a> being |alpha|.
-    
-    14. Add the key-value pair (|name| - |inputProperties|) to the <a>paint name to input properties
+
+    19. Add the key-value pair (|name| - |inputProperties|) to the <a>paint name to input properties
```
tabatkins: Use the Infra algos for manipulating a map? <https://infra.spec.whatwg.org/#map-set>

css-paint-api/Overview.bs
```diff
@@ -236,23 +278,32 @@ Paint Notation {#paint-notation}
 ================================
 
 <pre class='prod'>
-    <dfn>paint()</dfn> = paint( <<ident>> )
+    <dfn>paint()</dfn> = paint( <<ident>> [, <<custom-value>>]* )
```
tabatkins: `<<ident>>, <<custom-value>>#?`

Issue #378, opened 2017-04-07
[worklets] Provide hook for allowing downstream specs to specify a CSP destination.


bfgeek:
For example the css-paint-api spec should provide its worklet with a destination of "paintworklet".

Being more granular probably makes more sense here, instead of just a generic "worklet" destination.

bfgeek:
Sent https://github.com/whatwg/fetch/pull/527 for fetch integration.

bfgeek:
And sent https://github.com/w3c/webappsec-csp/pull/205 for CSP integration.

padenot:
What is preventing this to be merged, @bfgeek ? The two PRs above are merged.

bfgeek:
@padenot Done.

PR #379, opened 2017-04-07
[worklets] Fill in privacy and security sections of worklets.

### LINKED ISSUE #92: Worklets: Security considerations?
There's no "security considerations" section in the document, but I hope you'll consider writing one. 

Creating a new execution context in which JS executes is certainly interesting from a security standpoint. I'd like to see explicit discussion of the origin in which that code executes, and the privileges it ought to have. It would also be helpful to have a list of considerations that other authors should consider when determining whether a particular feature ought to be enabled for Worklets' "reduced API surface".

Likewise, how is the Worklet's script fetched? What Content Security Policy directive ought to apply? Should Worklets be available to non-secure contexts? Etc.

Thanks!

bfgeek:
Fixes #92. (better late than never? ;)

Broadly speaking worklets should be allowed in non-secure contexts as
downstream specs may want to use them there.

CSP wise this should work the same as workers, using the "child-src"
directive. I've filed issue #378 to allow each downstream spec to use a
unique destination, e.g. "paintworklet", "audioworklet", etc.

The CSP spec should probably be extended to have a "worklet-src"
directive (as there is now a "worker-src" directive now?).

### DIFF
--- a/worklets/Overview.bs
+++ b/worklets/Overview.bs
@@ -416,17 +416,12 @@ Worklets do not introduce any usable API surface to the web themselves, instead 
 specifications, e.g. [[css-paint-api-1]], [[webaudio]] will extend this specification to provide
 useful functionality.
 
-Worklets are allowed to be used outside <a>SecureContexts</a> as specifications which extend/use
-worklets may be allowed to used outside <a>SecureContexts</a>.
+Specifications which use worklets should decide if their worklet(s) should be allowed outside
+<a>SecureContexts</a>.
 
 Worklets load their module scripts in the same manner that workers load their module scripts. This
 fetches the top-level script by the <a>fetch a single module script</a> algorithm which sets the
 <a for=request>mode</a> to "<code>cors</code>".
-
-The behaviour of loading scripts can be restricted by using <a>child-src</a> [[CSP]] directive. Each
-specification which extends this specification <em>must</em> provide a <a
-for=request>destination</a> which is sent along with requests in the <a>Fetch a module worker script
-graph</a> algorithm.
 
 Issue(w3c/css-houdini-drafts#378): Provide hook for downstream specifications to provide their
     destination type.
--- a/worklets/Overview.bs
+++ b/worklets/Overview.bs
@@ -13,6 +13,7 @@ Editor: Ian Kilpatrick, ikilpatrick@chromium.org
 <pre class="anchors">
 urlPrefix: http://heycam.github.io/webidl/; type: dfn;
     text: AbortError
+    text: SecureContext
     text: SyntaxError
     urlPrefix: #idl-;
         text: DOMException
@@ -41,6 +42,7 @@ urlPrefix: https://html.spec.whatwg.org/multipage/webappapis.html; type: dfn;
     text: event loop
     text: event loop processing model
     text: fetch a module script tree
+    text: fetch a single module script
     text: global object
     text: https state
     text: incumbent settings object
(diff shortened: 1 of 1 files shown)


bfgeek:
@annevk @mikewest PTAL - :)

worklets/Overview.bs
```diff
@@ -410,7 +412,33 @@ loops and callbacks exceeding imposed time limits.
 Security Considerations {#security-considerations}
 ==================================================
 
-Issue(w3c/css-houdini-drafts#92): Need to decide if to allow worklets for unsecure context, etc.
+Worklets do not introduce any usable API surface to the web themselves, instead other
+specifications, e.g. [[css-paint-api-1]], [[webaudio]] will extend this specification to provide
+useful functionality.
+
+Worklets are allowed to be used outside <a>SecureContexts</a> as specifications which extend/use
+worklets may be allowed to used outside <a>SecureContexts</a>.
```
annevk: I'm not sure that's really good rationale. Ideally all new features just require secure contexts and certainly any specifications which wish to use worklets could do the same.

worklets/Overview.bs
```diff
@@ -410,7 +412,33 @@ loops and callbacks exceeding imposed time limits.
...
+
+Worklets load their module scripts in the same manner that workers load their module scripts. This
+fetches the top-level script by the <a>fetch a single module script</a> algorithm which sets the
+<a for=request>mode</a> to "<code>cors</code>".
+
+The behaviour of loading scripts can be restricted by using <a>child-src</a> [[CSP]] directive. Each
+specification which extends this specification <em>must</em> provide a <a
+for=request>destination</a> which is sent along with requests in the <a>Fetch a module worker script
+graph</a> algorithm.
```
annevk: Actually, worklets should be guarded by the script directive. But also, I'm not sure that needs to be mentioned in this section.

worklets/Overview.bs
bfgeek: Right - I don't think we have consensus as user agents yet that this is the case for all new features, which is why I worded it this way.

I've changed it to 
"Specifications which use worklets should decide if their worklet(s) should be allowed outside SecureContexts". instead?

worklets/Overview.bs
bfgeek: I've removed this section.

worklets/Overview.bs
```diff
@@ -410,7 +412,28 @@ loops and callbacks exceeding imposed time limits.
 Security Considerations {#security-considerations}
 ==================================================
 
-Issue(w3c/css-houdini-drafts#92): Need to decide if to allow worklets for unsecure context, etc.
+Worklets do not introduce any usable API surface to the web themselves, instead other
+specifications, e.g. [[css-paint-api-1]], [[webaudio]] will extend this specification to provide
+useful functionality.
+
+Specifications which use worklets should decide if their worklet(s) should be allowed outside
+<a>SecureContexts</a>.
```
annevk: I'd much rather we just limit them to secure contexts. No real reason to have new globals without it I think.

mikewest: @bfgeek: Have y'all thought about this feedback? Seeing the intent to ship threads for various worklet types without this fundamental question being answered worries me a bit.

bfgeek: I've bought this up on blink-dev. I linked to the HoudiniTF discussion there and there wasn't a clear conclusion.
https://lists.w3.org/Archives/Public/www-style/2017Mar/0023.html

I still err towards the side that it should be the decision of downstream specifications to decide this, but the concrete question that we need to answer at the moment is if the CSS Paint API should be available only on secure contexts or not.

annevk: I don't see why this should be decided downstream. It's a complex new feature and it can be trivially restricted. There's no good reason to expose it insecurely.

(As for the concerns around paint worklet. If there's no media query for secure contexts, just add one.)

ojanvafai: I think there are good reasons to expose it insecurely. But since the argument is not about CSS Paint specific concerns, but rather all new features, I commented on the relevant TAG issue. https://github.com/w3ctag/design-principles/pull/75#issuecomment-337647618

domenic:
@annevk, do you think we should try to port any of this over to HTML? Similarly, do you think there are other security concerns we should discuss, per https://github.com/w3c/css-houdini-drafts/issues/92 ?

annevk:
No, I don't think we should. These days I think security and privacy should be directly addressed in the relevant algorithms.

PR #381, opened 2017-04-11
[worklets] Fixes the setup of the inside setttings object for worklets.

### LINKED ISSUE #227: HTTPS state not defined for worklets
You say it comes from the global, but how did the global get it?

bfgeek:
Primarily based on:
https://html.spec.whatwg.org/#set-up-a-worker-environment-settings-object

Inherits a couple more things from the document.

Fixes #227 - inherits the HTTPS state from document

### DIFF
--- a/worklets/Overview.bs
+++ b/worklets/Overview.bs
@@ -26,56 +26,35 @@ Editor: Ian Kilpatrick, ikilpatrick@chromium.org
 }
 </style>
 
+<pre class=link-defaults>
+spec:fetch; type:dfn; for:/; text:fetch
+spec:html; type:dfn; for:/; text:browsing context
+spec:html; type:dfn; for:environment settings object;
+    text: global object
+    text: https state
+    text: referrer policy
+spec:html; type:dfn; for:environment; text:id
+spec:webidl; type:dfn; for:interface; text:inherit
+</pre>
+
 <pre class="anchors">
 urlPrefix: http://heycam.github.io/webidl/; type: dfn;
     text: AbortError
     text: SyntaxError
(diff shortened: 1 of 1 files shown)


bfgeek:
@annevk Can you take a look at this?

bfgeek:
...also cc/ @domenic

worklets/Overview.bs
```diff
@@ -265,24 +267,26 @@ When a user agent is to <dfn>set up a worklet environment settings object</dfn>,
...
 
         : The <a>API URL character encoding</a>
-        :: Return UTF-8.
+        :: Return <a>UTF-8</a>.
 
         : The <a>API base URL</a>
         :: Return |inheritedAPIBaseURL|.
 
-        : The <a>origin</a> and <a>effective script origin</a>
+        : The <a>origin</a>
```
annevk: You should probably use for="environment settings object" for this one.

worklets/Overview.bs
```diff
@@ -265,24 +267,26 @@ When a user agent is to <dfn>set up a worklet environment settings object</dfn>,
...
         :: Return |inheritedOrigin|.
```
annevk: Is there any reason to be same-origin with your creator? Might be safer to return an opaque origin. Though I guess that should be considered separately.

worklets/Overview.bs

worklets/Overview.bs
bfgeek: Yeah I just had a quick read through the html spec where it was used. It seems sane to me to use an opaque origin. Changed. We can always switch back to inherited if needed.

PR #386, opened 2017-04-13
[worklets] Reworks "Code Idempotency" section.

### LINKED ISSUE #308: [worklets] Make "Code Idempotency" section normative and stronger
https://drafts.css-houdini.org/worklets/#code-idempotency is a non-normative section that starts with:

> Multiple instances of WorkletGlobalScope can be created for each Worklet that they belong to. User agents may choose to do this in order to parallelize work over multiple threads, or to move work between threads as required.

In order to help make that possible, it goes on with content such as:

> authors […] should make their code idempotent.
> 
> User agents may choose to always have at least two WorkletGlobalScopes per Worklet

I think this is not enough. If one major implementation ships for example CSS Layout API _without_ these "may" precautions, it is likely that many websites will rely and worklet methods being executing in a certain order and in the same global scope. This might effectively make parallelization web-incompatible.

I’d like to make this section normative, and have it specify "must" requirements for user-agents. The details can be discussed, but one idea is to use JavaScript [Frozen Realms](https://github.com/FUDCo/proposal-frozen-realms) to make the global scope immutable (or have an immutable prototype so that creating a new global scope with that same prototype is cheap and could be done for every call).

CC @tschneidereit

### LINKED ISSUE #384: [worklets] "Multiple instances of WorkletGlobalScope can be created for each Worklet that they belong to."
This seems almost certainly wrong for audio worklets, in particular if those are going to use shared memory and such.

cc @bfgeek

bfgeek:
Fixes #384.

Fixes #308. (There is a separate issue for adding the 2
    WorkletGlobalScopes requirement for css-paint-api).

### DIFF
--- a/worklets/Overview.bs
+++ b/worklets/Overview.bs
@@ -99,28 +99,29 @@ expected to be shared between separate scripts. This is similar to the <a>docume
 Code Idempotency {#code-idempotency}
 ------------------------------------
 
-<em>This section is not normative.</em>
+Some specifications which use worklets ([[css-paint-api-1]]), allow user agents to parallelize work
+over multiple threads, or to move work between threads as required.
 
-Multiple instances of {{WorkletGlobalScope}} can be created for each {{Worklet}} that they belong
-to. User agents may choose to do this in order to parallelize work over multiple threads, or to move
-work between threads as required.
-
-Additionally different user agents may invoke a method on a class in a different order to other user
-agents.
+In these specifications user agents might invoke methods on a class in a different order to other
+user agents.
 
 As a result of this, to prevent this compatibility risk between user agents, authors who register
-classes on the global scope should make their code idempotent. That is, a method or set of methods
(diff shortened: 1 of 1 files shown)


bfgeek:
@annevk This should be fairly trivial.

This section could also just be moved to the downstream specs which require it, and removed this section; but its probably worth having this here just for context for folks reading it.

There is a separate issue for paint to firm update the global scope requirements.

annevk:
This looks good as an initial revision. I'd aim for something more formal down the line though, where a host specification defines a set of flags and based on that you allocate either one or two or more objects.

bfgeek:
Thats good feedback, I think we'll end up with two classes - one which only one global scope, the other type for layout/paint where we need this ideompotency requirement. A flag eventually makes sense.

SimonSapin:
CC @tschneidereit @metajack

PR #388, opened 2017-04-13
[worklets] Adds hook for downstream spec to set the destination.

### LINKED ISSUE #378: [worklets] Provide hook for allowing downstream specs to specify a CSP destination.

bfgeek:
Fixes #378.


bfgeek:
@annevk @mikewest This is mostly plumbing.

annevk:
You'll need to PR Fetch as well to add this destination. And I suspect this impacts CSP and maybe some other things too.

PR #389, opened 2017-04-14
[worklets] Introduce an "owner document" for each WorkletGlobalScope.

### LINKED ISSUE #385: [worklets] Owner of a worklet is not a formally defined thing
It's mentioned in passing that this could be a Window object, but there's no formal internal slot or definition surrounding this. We'll need one to fix https://github.com/whatwg/html/pull/2521 properly.

bfgeek:
Fixes #385.

### DIFF
--- a/worklets/Overview.bs
+++ b/worklets/Overview.bs
@@ -34,6 +34,7 @@ spec:html; type:dfn; for:environment settings object;
     text: https state
     text: referrer policy
 spec:html; type:dfn; for:environment; text:id
+spec:infra; type:dfn; text:list
 spec:webidl; type:dfn; for:interface; text:inherit
 </pre>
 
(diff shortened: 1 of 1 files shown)


bfgeek:
@annevk does this work?

annevk:
I guess I was not specific enough. The problem is mostly with WorkletGlobalScope. I see that you defer to host specifications for them lifetime, but I don't think that's good enough nor necessary. We know that all worklet globals will be owned by a Window/document so we might as well write that down.

bfgeek:
@annevk PTAL, I think this is enough? Each WorkletGlobalScope has an "owner document".

This also introduces a "destroy a WorkletGlobalScope" algorithm - may want to change the event loop in html spec for this instead...

annevk:
Do you have tests for this? In particular, if an initial about:blank document gets replaced, does the WorkletGlobalScope go away or stay? I'm asking since @wanderview wants to change ownership of workers to be global object based and maybe we want worklets to work like that too?

bfgeek:
We don't currently have tests for this; (@nhiroki we should add tests for this :) )

But we currently do kill the globals when the document get's destroyed, @nhiroki please correct me if I'm wrong:
https://cs.chromium.org/webrtc/src/third_party/WebKit/Source/core/workers/Worklet.cpp?q=worklet.cpp&l=65

nhiroki:
@bfgeek You're right.

I'm not sure how we can confirm it in web-platform-tests. For workers we could test it by creating a worker in a frame, closing the frame and postMessage+setTimeout like [1], but worklets don't have such a communication path.

[1] https://github.com/w3c/web-platform-tests/blob/master/workers/WorkerGlobalScope_close.htm

nhiroki:
Filed a chromium bug: https://crbug.com/712133

annevk:
@nhiroki with audio worklets it should be easy to confirm. Can a paint worklet draw anything that can then be imported by canvas or some such? Or does it intentionally not leak anything from its insides? (Sorry, I used to know this, but I forgot.)

nhiroki:
@annevk Thank you for the suggestion. I'll try the approach.

annevk:
It seems weird btw for the owner to not be the global object as you do expose the accessor to the worklet on the global object.

domenic:
With https://github.com/whatwg/html/pull/6056, we can close this. The lifetime of WorkletGlobalScopes is now very explicit.

PR #390, opened 2017-04-16
[css-paint-api] Removes invalidation based on size, and allows user-a…

### LINKED ISSUE #326: [css-paint-api] scope of paint-valid flag being box (not box+use/size) seems wrong
In the [section on Paint Invalidation](https://drafts.css-houdini.org/css-paint-api/#paint-invalidation) and the [section on Drawing an Image](https://drafts.css-houdini.org/css-paint-api/#drawing-an-image), the Painting API specification describes the concept of a paint-valid vs. paint-invalid flag.

It seems to describe this flag as being per-box, whereas I think it should either be:
- both per-box and per-use, or
- both per-box and per-concrete object size

In particular, if a style sheet has something like:

```
  background-image: paint(foo), paint(foo);
  background-size: 50px 50px, 100px 100px;
  background-position: top left, top right;
```

then the two images need to be painted separately, since they're different sizes, and the paint function would expect to be called separately for each size.  But if they were the same size, the implementation could presumably call the paint function only once (which is why I suggest per-concrete object size rather than per-use).  Though making it per-concrete object size requires changing a bit of the invalidation logic in the [section on Paint Invalidation](https://drafts.css-houdini.org/css-paint-api/#paint-invalidation).

There are similar problems for two uses of the same <code>paint()</code> function in different properties on the same element.

(I got here via https://github.com/w3ctag/spec-reviews/issues/140 .)

bfgeek:
…gents to cache.

Fixes #326.

I realized that explicitly placing an invalidation based on size here
was wrong, e.g. if a box changes size it has to run the "object size
negotiation" algorithm, which makes invalidating the paint() function
off size redundant.

Additionally allows user-agents to cache results from previous
invocations, and re-use. This allows UAs to re-use images between
instances if they have the same input style, concrete object size, and
input arguments.

### DIFF
--- a/css-paint-api/Overview.bs
+++ b/css-paint-api/Overview.bs
@@ -108,7 +108,7 @@ When the user agent wants to <dfn>invalidate paint functions</dfn> given |box|, 
 Performing <a>draw a paint image</a> results in the <a>paint valid flag</a> for a <<paint()>>
 function on a box to be set to <a>paint-valid</a>.
 
-<a>Invalidate paint functions</a> is expected to be run when the user agent recalculates computed
+<a>Invalidate paint functions</a> <em>must</em> be run when the user agent recalculates the computed
 style for a box.
 
 Note: In a future version of the spec, support could be added for partial invalidation. The user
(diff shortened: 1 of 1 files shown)


bfgeek:
@tabatkins @dbaron Does this make sense? We can talk about it at houdini tues.

dbaron:
Yeah, this looks good; when you described it in the meeting I didn't realize you'd added that additional wording.

The one minor comment is that "is expected to be run" should probably be a more normative phrase -- maybe **must**, although maybe **should** if there's a good reason for it not to be **must**.

bfgeek:
Thanks, changed.

PR #402, opened 2017-05-01
[worklets] Misc. edits from @domenic review.

### LINKED ISSUE #396: [worklets] Review
This is an overdue review of the latest changes, e.g. #375 and #381. I decided to just do a read-through of the algorithm portions of the spec, instead of those particularly. I've bolded the significant issues; most are minor.

- "Do not obtain any source texts for scripts or modules." is no longer necessary
- **"set up a worklet environment settings object" seems to inherit way too much:**
  - The API base URL is inherited from the HTML page? Meaning that URL resolution inside a worklet script created via `CSS.paintWorklet.loadModule("foo/bar.js")` should be relative to `index.html`, not relative to `foo/bar.js`?
  - The referrer policy ignores any HTTP headers for the worklet scripts, but instead just uses whatever the HTML page does?
  - The HTTPS state is copied from the document? So if I import a worklet over an insecure transport, it still counts as secure, as long as the hosting page uses HTTPS?
- Technically the id and creation URL for environment settings objects are fields, not algorithms, and so listing them separately might be better.
- Typo "consiting"
- In general we don't italicize "null"
- loadModule():
  - "the current Worklet" -> "this Worklet"
  - "SyntaxError" DOMException and other similar spec text should be using `"{{SyntaxError}}" {{DOMException}}` markup so it gets code-ified
  - Step 7 which establishes outsideSettings could move earlier, which would make step 3 which does URL parsing shorter
  - **"Run a module script" will try to report the exception, but WorkletGlobalScope doesn't have an onerror event handler**. Probably it should.
  - Great task-queuing discipline.

bfgeek:
Fixes #396.

### DIFF
--- a/worklets/Overview.bs
+++ b/worklets/Overview.bs
@@ -39,10 +39,6 @@ spec:webidl; type:dfn; for:interface; text:inherit
 
 <pre class="anchors">
 urlPrefix: http://heycam.github.io/webidl/; type: dfn;
-    text: AbortError
-    text: SyntaxError
-    urlPrefix: #idl-;
-        text: DOMException
     urlPrefix: #es-;
         text: invoking callback functions
 urlPrefix: https://html.spec.whatwg.org/multipage/workers.html; type: dfn;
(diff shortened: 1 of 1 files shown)


bfgeek:
@domenic :)

domenic:
LGTM, although I'm still concerned about the security implications of inheriting HTTPS state and referrer policy, so I'd appreciate either leaving #396 open for that or filing a separate issue.

bfgeek:
Thanks - I filed https://github.com/w3c/css-houdini-drafts/issues/403 to track separately.

Issue #276, opened 2016-08-10
[css-typed-om] Describe how the StylePropertyMaps should interact with custom properties


wilddamon:
Especially when using `getProperties()` and when using the iterators, and whether you should be able to set/get custom properties.

shans:
strawperson suggestion:

You should be able to get/set custom properties on inline and specified styles. You should be able to get custom properties on computed style.

getProperties:

* For inline style, should return the custom properties that are set on the inline style of the current node
* For specified style, should return the custom properties defined in the current rule
* For computed style, should return all defined custom properties on the current node (including inheritance)

shans:
(removing @apply because that isn't likely to happen any time soon)

tabatkins:
+1 to all three of those getProperties lines

tabatkins:
For typed custom properties with an initial value, they should show up in the computed style map.

tabatkins:
I went into more detail with this in 2519d4b and a preceding commit about the computed style.

PR #411, opened 2017-06-05
Fix for issue 276

nainar:
@shans could you take a look?

I don't know how to elaborate on the bit about warning authors. Couldn't find other examples in the Typed OM spec to mirror that language. 

Let me know. Thanks!


shans:
#276

shans:
This needs more detail, and direct descriptions rather than 'should' (which has normative meaning different to what we want).

PR #412, opened 2017-06-05
Fix issue 325 by adding algo for equals()

nainar:
@shans could you take a look, took a stab at writing up the algo for .equals()


css-typed-om/Overview.bs
```diff
@@ -784,10 +782,20 @@ The following are the arithmetic operations you can perform on dimensions:
...
 </div>
 
-<div class=issue>
-    Define {{CSSNumericValue/equals()}}.
+<div algorithm="CSSNumericValue.equals(...|value|)">
+    The <dfn for=CSSNumericValue method>equals(...|value|)</dfn> method,
+    when called on a {{CSSNumericValue}} |this|,
+    must perform the following steps:
+
+    1. If the |value| item is the same type as the {{CSSNumericValue}} the method is being called on continue to step 2, else return false.
```
shans: You don't need to continue on to step 2 as this happens automatically. Maybe rewrite as 'If |value| is of different type to |this|, return false.'

css-typed-om/Overview.bs
```diff
@@ -784,10 +782,20 @@ The following are the arithmetic operations you can perform on dimensions:
         is set to |this|.
 </div>
 
-<div class=issue>
-    Define {{CSSNumericValue/equals()}}.
+<div algorithm="CSSNumericValue.equals(...|value|)">
+    The <dfn for=CSSNumericValue method>equals(...|value|)</dfn> method,
+    when called on a {{CSSNumericValue}} |this|,
+    must perform the following steps:
+
```
shans: Will need to either explicitly iterate over |value| (change to values?) or just restrict this to take a single value.

css-typed-om/Overview.bs
shans: Also we need to define what 'different type' means.

css-typed-om/Overview.bs
```diff
@@ -784,10 +782,20 @@ The following are the arithmetic operations you can perform on dimensions:
...
+    1. If the |value| item is the same type as the {{CSSNumericValue}} the method is being called on continue to step 2, else return false.
+
+    2. If the |value| item has the same type as the {{CSSNumericValue}} continue to the next step, else return false.
```
shans: Isn't this the same as step 1?

css-typed-om/Overview.bs
```diff
@@ -784,10 +782,20 @@ The following are the arithmetic operations you can perform on dimensions:
...
+
+    3. If the |value| item has the same unit as the {{CSSNumericValue}} return true, else return false.
+
```
shans: Also need to handle calc values, etc.

css-typed-om/Overview.bs
nainar: This was supposed to be class.

css-typed-om/Overview.bs
```diff
@@ -784,10 +782,20 @@ The following are the arithmetic operations you can perform on dimensions:
         is set to |this|.
 </div>
 
-<div class=issue>
-    Define {{CSSNumericValue/equals()}}.
+<div algorithm="CSSNumericValue.equals(...|value|)">
+    The <dfn for=CSSNumericValue method>equals(...|value|)</dfn> method,
```
tabatkins: Note that .equals() is variadic; you have to check that *all* the values are equal to each other/this.

css-typed-om/Overview.bs
tabatkins: * Just reverse this conditional and return false. No need to explicitly say to go to the next step; that's implied by the list.  
* Be clear what you mean by "type" - you use the exact same word in the second step, but clearly mean something different.  I suspect here you're talking about the interface type.
* "the CSSNumericValue the method is being called on" is already stored in the `|this|` variable, so you just refer to it that way.

css-typed-om/Overview.bs
tabatkins: * Same concerns as about the structure; reverse the conditional and just return false.
* I think you mean "value" here? Refer to `|this|’s {{CSSUnitValue/value}} internal slot` then (and same for |value|).

css-typed-om/Overview.bs
```diff
@@ -784,10 +782,20 @@ The following are the arithmetic operations you can perform on dimensions:
...
+    when called on a {{CSSNumericValue}} |this|,
+    must perform the following steps:
+
+    1. If the |value| item is the same type as the {{CSSNumericValue}} the method is being called on continue to step 2, else return false.
+
+    2. If the |value| item has the same type as the {{CSSNumericValue}} continue to the next step, else return false.
+
+    3. If the |value| item has the same unit as the {{CSSNumericValue}} return true, else return false.
```
tabatkins: Use `{{CSSUnitValue/unit}} internal slot`.

css-typed-om/Overview.bs
```diff
@@ -784,10 +782,20 @@ The following are the arithmetic operations you can perform on dimensions:
         is set to |this|.
 </div>
 
-<div class=issue>
-    Define {{CSSNumericValue/equals()}}.
+<div algorithm="CSSNumericValue.equals(...|value|)">
+    The <dfn for=CSSNumericValue method>equals(...|value|)</dfn> method,
+    when called on a {{CSSNumericValue}} |this|,
+    must perform the following steps:
```
tabatkins: The algo claims to be defined for CSSNumericValue, but the steps assume that you're looking at a CSSUnitValue.  It needs to work for all the CSSMathValue subtypes too.

tabatkins:
Whoops, I still had this page up from earlier, and it hadn't updated with Shane's comments. Sorry for doubling up. ^_^

css-typed-om/Overview.bs
tabatkins: And by "or" Shane means "do the first thing", because all the other methods take multiple values too.

nainar:
I suck at Github but I have added a new commit. Hopefully that addresses both @shans and @tabatkins comments.

css-typed-om/Overview.bs
nainar: Done in commit 3

css-typed-om/Overview.bs
nainar: Had done the first thing. Commit 3 should make equals accept an array.

css-typed-om/Overview.bs
nainar: Done in commit 3

css-typed-om/Overview.bs
nainar: Done in commit 3

css-typed-om/Overview.bs
nainar: Done in commit 3

css-typed-om/Overview.bs
nainar: Done in commit 3

css-typed-om/Overview.bs
```diff
@@ -784,10 +782,29 @@ The following are the arithmetic operations you can perform on dimensions:
...
+
+    1. Replace each [=list/item=] of |values|
+        with the result of [=rectifying a numberish value=] for the [=list/item=].
+
+    2. Let |thisAndValues| be |this| prepended to |values|.
+
+    1. [=list/If all=] |item|s of |thisAndValues| do not belong to the same class return false. Example: all items in |thisAndValues| must be {{CSSUnitValue}}.
+
+    2. [=list/If all=] |item|s in |thisAndValues| are {{CSSUnitValue}s and have the same {{CSSUnitValue/value}} return true.
```
tabatkins: unit *and* value ^_^

css-typed-om/Overview.bs
```diff
@@ -784,10 +782,29 @@ The following are the arithmetic operations you can perform on dimensions:
...
+
+    4. =list/If all=] |item|s in |thisAndValues| are {{CSSMathValue}s :
```
tabatkins: Missing `[`

css-typed-om/Overview.bs
```diff
@@ -784,10 +782,29 @@ The following are the arithmetic operations you can perform on dimensions:
...
+        1. If the {{CSSMathValue/operator}} are not equal for all {{CSSMathValues}} in |thisAndValues| return false.
```
tabatkins: This is already covered by "same class" earlier, so you can drop this.

css-typed-om/Overview.bs
```diff
@@ -784,10 +782,29 @@ The following are the arithmetic operations you can perform on dimensions:
...
+        2. If |thisAndValues| are all either {{CSSMathSum}}, {{CSSMathProduct}}, {{CSSMathMin}} or {{CSSMathMax}}, iterate over the {{CSSNumericArray}}s on |thisAndValues|,
+        	1. For each {{CSSUnitValue}} in each of the {{CSSNumericArray}}s go back to step 1 to check that the {{CSSUnitValue}}s on all items in |thisAndValue| are equal.
+    	3. If |this| and |value| are both either {{CSSMathNegate}} or {{CSSMathInvert}}:
```
tabatkins: Need to handle multiples.

nainar:
@tabatkins took a stab at equals take a look?

css-typed-om/Overview.bs
nainar: Done

css-typed-om/Overview.bs
nainar: Done

css-typed-om/Overview.bs
nainar: Done

css-typed-om/Overview.bs
nainar: Do you mean CSSMathProduct? Cause that is dealt with in the above step. Or do you mean the case where values in |thisAndValues| are not all of the same type? If the latter in that case we should return a false right?

Issue #404, opened 2017-05-09
[css-typed-om] should have min() and max() functions on CSSNumericValue


shans:
(to sit alongside add(), sub(), mul(), div()).

tabatkins:
Fixed by #413.

PR #413, opened 2017-06-05
Fix issue 404

nainar:
@shans could you take a look?

This hopefully adds the text needed for issue #404

### DIFF
--- a/css-typed-om/Overview.bs
+++ b/css-typed-om/Overview.bs
@@ -595,6 +595,9 @@ are represented by subclasses of the {{CSSNumericValue}} interface.
         CSSNumericValue mul(CSSNumberish... values);
         CSSNumericValue div(CSSNumberish... values);
 
+        CSSNumericValue min(CSSNumberish... values);
+        CSSNumericValue max(CSSNumberish... values);
+
         boolean equals(CSSNumberish... value);
 
         CSSNumericValue to(DOMString unit);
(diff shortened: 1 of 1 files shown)


css-typed-om/Overview.bs
```diff
@@ -784,6 +787,72 @@ The following are the arithmetic operations you can perform on dimensions:
...
+
+    4. If all of the [=list/items=] in |values| are {{CSSUnitValue}}s
+        and have the same {{CSSUnitValue/unit}},
+        return a new {{CSSUnitValue}}
+        whose {{CSSUnitValue/unit}} internal slot is set to |this|’s {{CSSUnitValue/unit}} internal slot,
+        and {{CSSUnitValue/value}} internal slot is set to the minimum value of
+        the {{CSSUnitValue/value}} internal slots
+        of the [=list/items=] in |values|.
+
+    5. If |this| is a {{CSSMathMin}} object,
```
tabatkins: Oh shoot, you're right, I didn't intend to do an in-place modification in the add() function. My intention is just that, if you already have a Sum, we should just create a larger Sum with all the values, rather than nesting the Sums.

That is, `CSS.num(1).add(1).add(2)` should give the same result as `CSS.num(1).add(2, 3)`.

css-typed-om/Overview.bs
```diff
@@ -784,6 +787,72 @@ The following are the arithmetic operations you can perform on dimensions:
...
+    must perform the following steps:
+
+    1. Replace each [=list/item=] of |values|
+        with the result of [=rectifying a numberish value=] for the [=list/item=].
+
+    2. Prepend |this| to |values|.
+
+    3. If there are at least two different [=strong types=]
+        among the [=list/items=] in |values|,
+        [=throw=] a {{TypeError}}.
```
tabatkins: Throwing a TypeError is correct. You can't compare a length to an angle; doing so will be rejected as a syntax error in CSS, and throws a TypeError here in JS.

css-typed-om/Overview.bs
tabatkins: That said, go ahead and stay consistent for this; I need to rewrite the math functions now to take advantage of the proper type system. :(

tabatkins: And I've done that now, so instead be consistent with text of the new functions. ^_^

nainar:
@tabatkins take a look please?

css-typed-om/Overview.bs
```diff
@@ -717,6 +720,52 @@ The following are the arithmetic operations you can perform on dimensions:
...
+        [=list/prepend=] the [=list/items=] in |this|’s {{CSSMathMin/values}} internal slot
+        to |values|.
+        Otherwise,
+        prepend |this| to |values|.
+
+    3. Let |type| be the result of [=adding=] the [=types=] of every [=list/item=] in |values|.
+        If |type| is failure,
+        [=throw=] a {{TypeError}}.
+
+    4. Return a new {{CSSMathMin}} object
```
tabatkins: Max. ^_^ (Several places in here.)

nainar: /facepalm.

Done.

PR #419, opened 2017-06-24
[css-properties-values-api] remove duplicated "&lt;"

hwanseung:
remove duplicated "&lt;" in css-properties-values-api.


hwanseung:
@shans could you review this patch?

Issue #354, opened 2017-02-03
[css-properties-and-values] Setting/registering is allowed to be order-dependent


tabatkins:
Per [Seattle resolution](https://log.csswg.org/irc.w3.org/houdini/2017-01-10/#e756205) (resolution is badly worded, read up to figure out what was actually being discussed), we're OK with an order-dependence between setting a value and registering a value.

That is, we want the following behavior:

```js
CSS.registerProperty("--foo", { grammar: "<color>" });
el.styleMap.set("--foo", "red");
el.styleMap.set("--foo", "hello"); // throws an error
// --foo is still set to "red"
```

and 

```js
el.styleMap.set("--foo", "red");
el.styleMap.set("--foo", "hello"); // valid at this point; --foo is untyped
CSS.registerProperty("--foo", { grammar: "<color>" }); // valid
// because --foo was set to a non-<color> value,
// it's now set to "unset", the bottom value for properties
```

tabatkins:
Further clarification: per [Seattle resolution](https://log.csswg.org/irc.w3.org/houdini/2017-01-10/#e756301), when using a typed property *in a stylesheet*, setting to an invalid value doesn't trigger fallback.  That is:

```
<script>
CSS.registerProperty("--foo", {grammar: "<color>"});
</script>
<style>
div { --foo: red; }
div.foo {  --foo: hello; }
</style>
```

In here, the "--foo: hello" is *not* thrown out at parse-time; it's accepted, but because it's invalid, it gets treated as "--foo: unset;".

* This is an intentional break from "real" CSS properties, because the threat model is different (unpredictable browsers with varying support, vs known libraries with well-known support).  
* This *removes* the order-dependence between setting and registration when you're setting in a stylesheet, rather than thru the JS API.
* This makes typed custom properties less consistent with "real" CSS properties, but more consistent with *untyped* custom properties.

shans:
What specifically needs to change in Typed OM for this? Seems like mainly a properties & values thing?

Issue #321, opened 2016-10-06
[css-properties-values-api] Substituting registered properties with var()


tim-loh:
Are registered properties substituted as lists of tokens as per css-variables or as some sort of parsed representation? There are some cases where values can be parsed as different types so this affects whether the end value parses successfully, e.g.

--number: 0; height: var(--number);
--color: green; animation-name: var(--color);
--length-list: 5px 10px 20px; box-shadow: green var(--length-list);
--url: url(a.png); background-image: var(--url); /\* &lt;image> values can be &lt;url> values so this should be valid */

For the last case, if the declarations are in separate files, do we resolve &lt;url>/&lt;image> types with a base URL from the sheet with variable declaration or the usage thereof? When these are not registered, it seems that we should use the "background-image" sheet (crbug.com/618165#c11) but doing this for a registered custom property means the custom property has a different computed value to where it's used.

shans:
(nit: as of TPAC, unregistered properties are substituted as strings, not lists of tokens).

It's interesting that there are two truthy-feeling principles opposed here:

1. the computed value of a registered custom property shouldn't differ from the computed value part of  a property that references that custom property (when the types align)
2. registering a custom property shouldn't change its behavior
 
(1 means roughly that you shouldn't ever be able to have
```
--foo: [something];
property-with-type-matching-foo: var(--foo);
```
and `getComputedStyle(el).getPropertyValue('--foo') !==  getComputedStyle(el).getPropertyValue('property-with-type-matching-foo');` )

I don't really know what to do here. It does seem deeply weird for the variable to not resolve relative to the variable sheet when you know up-front that it is a url().

tabatkins:
> (nit: as of TPAC, unregistered properties are substituted as strings, not lists of tokens).

No, this is inaccurate. (It would be *crazy wrong* taken literally as you said.) Untyped properties have their values *preserved* as literal strings, such that their serialization is identical to their input.  They're definitely *substituted* as token streams, tho; otherwise you get very quirky parsing things where a single token can span across two var() substitutions.

Typed properties actually have typed values, tho, and get substituted as such. So your first and second examples should both be errors (height doesn't take numbers, animation-name doesn't take colors), your third should work as long as it's a correct number of lengths, and your fourth is fine.

[...]

PR #427, opened 2017-07-19
[css-properties-values-api] Typed custom property declarations are syntax-checked at computed-value time, not parse time.

jyc:
Clarify that the `syntax` field is used during computed-value time,
write in the section about calculation of computed values that
declarations can be found invalid at computed-value time, and update the
example which previously indicated syntax-checking at parse time.

Per Seattle resolution, as expressed in the following comments:

> The original issue (should a property fail at parse-time if a typed
> custom property is used in a wrong way, per the property's grammar) is,
> I think, a no. I believe this falls out of our decision at the Seattle
> f2f (#354) that registering a custom property shouldn't require an
> immediate global re-parse. You can do wrong things to/with a typed
> custom property; it'll just cause a failure at computed-value time, like
> an untyped custom property with an incompatible value does.
>
> - https://github.com/w3c/css-houdini-drafts/issues/321#issuecomment-279794319

> Further clarification: per Seattle resolution, when using a typed
> property in a stylesheet, setting to an invalid value doesn't trigger
> fallback ...
>
> - https://github.com/w3c/css-houdini-drafts/issues/354#issuecomment-277341782

### DIFF
--- a/css-properties-values-api/Overview.bs
+++ b/css-properties-values-api/Overview.bs
@@ -92,7 +92,7 @@ following members:
 ::  The name of the custom property being defined.
 
 :   <dfn dict-member for=PropertyDescriptor>syntax</dfn>
-::  A string representing how this custom property is parsed.
+::  A string representing how this custom property is parsed to obtain its computed value.
 
 :   <dfn dict-member for=PropertyDescriptor>inherits</dfn>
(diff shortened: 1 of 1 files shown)


dholbert:
@tabatkins , perhaps you could review this pull request (or suggest another Properties & Values spec-editor to review it)?

Issue #149, opened 2016-04-04
[css-typed-om] Describe that StylePropertyMaps are not live objects


nainar:

nainar:
@tabatkins

jyc:
What does this mean? If a CSS declaration or inline style is otherwise modified (say, through JS), does the specified or inline StylePropertyMap update? What happens if we then call StylePropertyMap::append, for example? Do we append to the declaration or inline style based on its current value while appending to the StylePropertyMap's value based on what it was created with? What does the iterator give us for that property?

jyc:
Found a definition of "live object" here: https://dom.spec.whatwg.org/#concept-collection-live .
Is this the correct definition? If so, and these are not live objects and therefore do not "operate on the actual underlying data, not a snapshot of the data," what is the point of mutable StylePropertyMpas?

shans:
To clarify, this is saying that StylePropertyMaps won't update based on style changes from other sources:

```
var map = el.styleMap;
map.get('height') // for argument's sake, CSSSimpleLength(100, 'px')
el.style.height = '200px';
map.get('height') // will still return CSSSimpleLength(100, 'px')
```

It might actually be reasonably simple to support live maps for specified objects but for computed styles this gets quite expensive, and it seems best to be consistent.

shans:
feedback from jyc on irc: this should probably be true only for computed maps, not specified maps.

tabatkins:
Ah, that's probably fine. I was getting a little confused why .set() worked but wasn't live - there's clearly still a communication channel there.  (This doesn't really need to be "live", per se, it's just that .get() goes and gets the value off of the element.)

So yeah, the Readonly version is non-live.

dbaron:
Resolved to be live [for specified](https://log.csswg.org/irc.w3.org/houdini/2016-09-22/#e727261) and [for computed](https://log.csswg.org/irc.w3.org/houdini/2016-09-22/#e727285) style maps.

shans:
The only map that's left is the inline map. We didn't specifically talk about this during the spetember f2f but it seems it should be live too. OK for me to close this issue as invalid?

PR #431, opened 2017-07-25
StylePropertyMaps are live objects.

jm318:
This change is to address issue 149 (https://github.com/w3c/css-houdini-drafts/issues/149). The conclusion  reached at W3C meeting was that computed, declared and inline StylePropertyMap should all be live objects. This change puts in the conclusion.


jm318:
@shans PTAL. Thanks!

PR #439, opened 2017-08-01
Allow speculative evaluation

### LINKED ISSUE #406: [worklets] Speculative evaluation
In the "code idempotency" section of the draft spec (https://drafts.css-houdini.org/worklets/#code-idempotency) there's an encouragement to write idempotent code, because UAs might swap globals.

UAs may also wish to speculatively evaluate worklets: for example in the paint worklet, the calculated values for CSS may be available before the concrete object size, so a UA might speculatively evaluate the worklet using an old value for the size, and discard the result if the speculation turns out to be wrong.

It might be nice if the spec explicitly allowed this?

asajeffrey:
Currently, user agents can only invoke worklets with consistent snapshots of the state. This PR allows user agents to invoke worklet scripts with inconsistent state, as long as they discard the result.

The particular instance of this is paint worklets, since the CSS style values are known before the concrete size, so in Servo we speculate that the size hasn't changed (https://github.com/servo/servo/pull/17810).

Fixes #406.

### DIFF
--- a/css-paint-api/Overview.bs
+++ b/css-paint-api/Overview.bs
@@ -451,9 +451,8 @@ Note: In a future version of the spec, the author could have the ability to spec
 The {{PaintSize}} object represents the size of the image that the author should draw. This is
 the <a>concrete object size</a> given by the user agent.
 
-The <a>draw a paint image</a> function may be speculatively invoked by
-the user agent at any point, with any |concreteObjectSize|. The
-resulting image is not displayed.
+The <a>draw a paint image</a> function may be speculatively invoked by the user agent at any point,
+with any |concreteObjectSize|. The resulting image is not displayed.
 
 Note: User agents may use any heuristic to speculate a possible future value
     for |concretObjectSize|, for example speculating that the size remains unchanged.
--- a/css-paint-api/Overview.bs
+++ b/css-paint-api/Overview.bs
@@ -437,7 +437,8 @@ Note: The user agent can optionally defer drawing images which are <a>paint-inva
 </div>
 
 The <a>draw a paint image</a> function is invoked by the user agent during the <a>object size
-negotiation</a> algorithm which is responsible for rendering an <<image>>.
+negotiation</a> algorithm which is responsible for rendering an <<image>>,
+with |concreteObjectSize| set to the <a>concrete object size</a> of the <a>box</a>.
 
 For the purposes of the <a>object size negotiation</a> algorithm, the paint image has no
 <a>intrinsic dimensions</a>.
@@ -450,6 +451,13 @@ Note: In a future version of the spec, the author could have the ability to spec
 The {{PaintSize}} object represents the size of the image that the author should draw. This is
 the <a>concrete object size</a> given by the user agent.
 
+The <a>draw a paint image</a> function may be speculatively invoked by
+the user agent at any point, with any |concreteObjectSize|. The
+resulting image is discarded.
+
+Note: Although the image is discarded, it may still be cached, so subsequent invocations
+    of <<paint()>> may use the cached image.
+
 <pre class='idl'>
 [Exposed=PaintWorklet]
 interface PaintSize {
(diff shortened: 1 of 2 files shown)


css-paint-api/Overview.bs
```diff
@@ -461,12 +469,12 @@ interface PaintSize {
...
 into its appropriate stacking level (as defined by the property the CSS property its associated
-with), given its |concreteObjectSize| (<a>concrete object size</a>) it <em>must</em> run the
-following steps:
+with), given |concreteObjectSize| it <em>must</em> run the following steps:
     1. Let |paintFunction| be the <<paint()>> function on the |box| which the user agent wants to
         draw.
 
-    2. If the <a>paint valid flag</a> for the |paintFunction| is <a>paint-valid</a> the user agent
+    2. If the <a>paint valid flag</a> for the |paintFunction| is <a>paint-valid</a>,
+        and the previous invocation had the same |concreteObjectSize|, the user agent
```
tschneidereit: This implies that there can only ever be one speculative evaluation at a time. If we want to allow speculating multiple frames ahead, this should probably say something about checking for results of a speculative evaluation that match the input state.

css-paint-api/Overview.bs
```diff
@@ -450,6 +451,13 @@ Note: In a future version of the spec, the author could have the ability to spec
 The {{PaintSize}} object represents the size of the image that the author should draw. This is
 the <a>concrete object size</a> given by the user agent.
 
+The <a>draw a paint image</a> function may be speculatively invoked by
+the user agent at any point, with any |concreteObjectSize|. The
+resulting image is discarded.
+
+Note: Although the image is discarded, it may still be cached, so subsequent invocations
+    of <<paint()>> may use the cached image.
+
```
asajeffrey: I added text to the note: "Note: User agents may use whichever heuristics to speculate a possible future value for |concretObjectSize|, for example speculating that the size remains unchanged." I'm not sure that we want to say more about how to speculate in the normative text.

I replaced "discarded" by "not displayed", which is hopefully less confusing.

css-paint-api/Overview.bs
asajeffrey: All the discussion of the paint valid flag is dubious, perhaps we should just delete it? It's very single-threaded. This is probably a separate issue though.

worklets/Overview.bs
```diff
@@ -119,6 +119,30 @@ The following techniques are used in order to encourage authors to write code in
...
+invoke methods on a class based on the state of the user agent. To
+increase the concurrency between threads, a user agent may invoke a
+method speculatively, before a consistent snapshot of the state is
+known.
+
+In these specifications user agents might invoke methods on a class at
+any time, and with any arguments, not just ones corresponding to
+consistent states of the user agent. The results of such speculative
+evaluations will be discarded, but may be cached for later use, thus
+increasing the concurrency between the user agent and worklet threads.
```
asajeffrey: Good point, I tried rephrasing it.

css-paint-api/Overview.bs
bfgeek: There is a separate section which allows for the caching inside "invoke a paint callback", see step 10.

worklets/Overview.bs
```diff
@@ -119,6 +119,30 @@ The following techniques are used in order to encourage authors to write code in
 
  - User agents can create and destroy {{WorkletGlobalScope}}s at any time for these specifications.
 
+Speculative Evaluation {#speculative-evaluation}
+------------------------------------------------
+
+Some specifications which use worklets ([[css-paint-api-1]]) may
```
bfgeek: please reflow to 100 colwidth.

css-paint-api/Overview.bs
```diff
@@ -450,6 +451,16 @@ Note: In a future version of the spec, the author could have the ability to spec
 The {{PaintSize}} object represents the size of the image that the author should draw. This is
 the <a>concrete object size</a> given by the user agent.
 
+The <a>draw a paint image</a> function may be speculatively invoked by
```
bfgeek: please reflow to 100 colwidth.

css-paint-api/Overview.bs
asajeffrey: @bfgeek yes, this is why I was thinking we could just remove all the stuff about the paint valid flag, since we're already allowing caching inside "invoke a paint callback". Shall I submit a separate PR for that?

bfgeek: Yeah that sounds good, we still need to describe how the inputProperties invalidate that image on a box, but we can have that in prose vs. as an algorithm.

asajeffrey: OK, I'll file an issue for this.

css-paint-api/Overview.bs
asajeffrey: Done.

worklets/Overview.bs
asajeffrey: Done.

asajeffrey:
Thanks!

Issue #148, opened 2016-04-04
[css-typed-om] add detailed descriptions of the rest of the methods on StylePropertyMap


nainar:

nainar:
@tabatkins

shans:
This comprises the following list:
* get a value from a StylePropertyMap
* set a value on a StylePropertyMap
* getAll
* has
* delete

nainar:
The last one in this list was pull request 451. Closing this issue now.

PR #442, opened 2017-08-03
Add description for StylePropertyMapReadOnly.has()

nainar:
@shans PTAL? This adds a description for has() which is one of the functions mentioned in #148.

Will do the other functions in different CLs.


shans:
Probably should wait for the get() text to be merged before merging this though.

nainar:
Thank you - will merge it in when get() gets merged :)

Issue #436, opened 2017-08-01
[css-paint-api] Consider adding property for expected rendered bitmap size.


flackr:
The rendering context in https://drafts.css-houdini.org/css-paint-api-1/#2d-rendering-context defines the output bitmap size to be the size of the fragment being rendered however the actual size of the rendered bitmap used internally may vary.

In order to allow output density sensitive painting we may want to expose the size of the rendered bitmap we intend to use (or an approximation) so that the author may produce a better rendering for that output size. The author may also use this to reduce the quality of their rendering given an extremely low density (i.e. preview sized renderings may not need to paint as much) or to paint hairline lines.

Options:
1. Expose rasterBackingScale on canvas context
2. Pass additional parameter to paint.

bfgeek:
The `devicePixelRatio` has been added on the `PaintRenderingContext2D` in the above commit.

PR #446, opened 2017-08-07
Add devicePixelRatio property to PaintRenderingContext2D

xidachen:
Per discussion here: https://github.com/w3c/css-houdini-drafts/issues/436, CSS working group agreed to add a devicePixelRatio property to the PaintRenderingContext2D.

@bfgeek Please review.

### DIFF
--- a/css-paint-api/Overview.bs
+++ b/css-paint-api/Overview.bs
@@ -131,10 +131,10 @@ partial interface CSS {
 
 The {{PaintWorkletGlobalScope}} is the global execution context of the {{paintWorklet}}.
 
-The {{PaintWorkletGlobalScope}} has a devicePixelRatio property, which represents the ratio
-of the resolution in physical pixels to the resolution of CSS pixels for the current display
-device. When measured from the same display device, this value must be identical to the
-Window.devicePixelRatio.
+The {{PaintWorkletGlobalScope}} has a {{PaintWorkletGlobalScope/devicePixelRatio}} property,
+which represents the ratio of the resolution in physical pixels to the resolution of CSS
+pixels for the current display device. When measured from the same display device, this
+value must be identical to the Window.devicePixelRatio.
 
 <pre class='idl'>
 [Global=(Worklet,PaintWorklet),Exposed=PaintWorklet]
--- a/css-paint-api/Overview.bs
+++ b/css-paint-api/Overview.bs
@@ -131,10 +131,8 @@ partial interface CSS {
 
 The {{PaintWorkletGlobalScope}} is the global execution context of the {{paintWorklet}}.
 
-The {{PaintWorkletGlobalScope}} has a {{PaintWorkletGlobalScope/devicePixelRatio}} property,
-which represents the ratio of the resolution in physical pixels to the resolution of CSS
-pixels for the current display device. When measured from the same display device, this
-value must be identical to the Window.{{Window/devicePixelRatio}}.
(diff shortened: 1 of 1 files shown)


flackr:
No, in the discussion we resolved on adding it to the PaintWorkletGlobalScope so it would be consistent with devicePixelRatio.

xidachen:
I looked at the discussion history and you are right. I will make changes accordingly. Thanks.

xidachen:
@flackr @bfgeek Please review the new commit. Thank you.

css-paint-api/Overview.bs
```diff
@@ -131,10 +131,16 @@ partial interface CSS {
 
 The {{PaintWorkletGlobalScope}} is the global execution context of the {{paintWorklet}}.
 
+The {{PaintWorkletGlobalScope}} has a devicePixelRatio property, which represents the ratio
```
bfgeek: You can link to devicePixelRatio via. {{PaintWorkletGlobalScope/devicePixelRatio}}

xidachen:
@bfgeek added link to devicePixelRatio

css-paint-api/Overview.bs
```diff
@@ -131,10 +131,16 @@ partial interface CSS {
 
 The {{PaintWorkletGlobalScope}} is the global execution context of the {{paintWorklet}}.
 
+The {{PaintWorkletGlobalScope}} has a {{PaintWorkletGlobalScope/devicePixelRatio}} property,
+which represents the ratio of the resolution in physical pixels to the resolution of CSS
+pixels for the current display device. When measured from the same display device, this
+value must be identical to the Window.devicePixelRatio.
```
flackr: I was going to ask to reference devicePixelRatio, but it looks to me like the csswg has [resolved not to standardize devicePixelRatio](https://github.com/w3c/csswg-drafts/issues/417#issuecomment-267111094), see also this blog post for [unprefixing -webkit-device-pixel-ratio](https://www.w3.org/blog/CSS/2012/06/14/unprefix-webkit-device-pixel-ratio/). If the working group hasn't changed their stance on this, perhaps it's worth a note that we should aim to track whatever the expected way of querying the display density from javascript is. That thread suggests that the standard way is resolution but as far as I can tell you're unable to query that from javascript.

xidachen:
@flackr Please review. The new commit now reference to {{Window/devicePixelRatio}}.

css-paint-api/Overview.bs
```diff
@@ -131,10 +131,16 @@ partial interface CSS {
 
 The {{PaintWorkletGlobalScope}} is the global execution context of the {{paintWorklet}}.
 
+The {{PaintWorkletGlobalScope}} has a {{PaintWorkletGlobalScope/devicePixelRatio}} property,
+which represents the ratio of the resolution in physical pixels to the resolution of CSS
+pixels for the current display device. When measured from the same display device, this
+value must be identical to the Window.{{Window/devicePixelRatio}}.
```
flackr: Since we link to the spec which has a more precise definition of the device pixel ratio it probably makes sense to shorten this and not try to redefine what it means here. i.e. How about just The PaintWorkletGlobalScope has a devicePixelRatio property which is identical to the Window.devicePixelRatio property.

xidachen:
@flackr Thanks. I made the change, it is much simpler now :)

xidachen:
@bfgeek This can be merged

PR #448, opened 2017-08-17
Replace alpha flag with a PaintRenderingContext2DSettings object

xidachen:
@bfgeek @flackr: Please review.

### DIFF
--- a/css-paint-api/Overview.bs
+++ b/css-paint-api/Overview.bs
@@ -138,11 +138,11 @@ interface PaintWorkletGlobalScope : WorkletGlobalScope {
 };
 </pre>
 
-The {{ContextAttributes}} allows the settings for the rendering context associated
-with the paint canvas. Right now, the {{ContextAttributes}} has a boolean flag
+The {{PaintRenderingContext2DSettings}} allows the settings for the rendering context associated
+with the paint canvas. Right now, the {{PaintRenderingContext2DSettings}} has a boolean flag
 indicating if the paint canvas contains an alpha channel.
 <pre class='idl'>
-dictionary ContextAttributes {
+dictionary PaintRenderingContext2DSettings {
     boolean alpha = true;
 };
 </pre>
@@ -153,7 +153,7 @@ dictionary ContextAttributes {
         class MyPaint {
             static get inputProperties() { return ['--foo']; }
             static get inputArguments() { return ['&lt;color>']; }
-            static get contextAttributes() { return {alpha: true}; }
+            static get paintRenderingContext2DSettings() { return {alpha: true}; }
 
             paint(ctx, size, styleMap) {
                 // Paint code goes here.
@@ -178,7 +178,7 @@ A <dfn>paint definition</dfn> is a <a>struct</a> which describes the information
  - <dfn for="paint definition">input properties</dfn> which is a <a>list</a> of
     <code>DOMStrings</code>.
 
- - A <dfn for="paint definition">ContextAttributes object</dfn>.
+ - A <dfn for="paint definition">PaintRenderingContext2DSettings object</dfn>.
 
 A <dfn>document paint definition</dfn> is a <a>struct</a> which describes the information
 needed by the <a>document</a> about the author defined <<image>> function (which can be referenced
@@ -190,7 +190,7 @@ by the paint function). It consists of:
  - A <dfn for="document paint definition">input argument syntaxes</dfn> which is a <a>list</a> of
     parsed [[css-properties-values-api-1#supported-syntax-strings]].
 
- - A <dfn for="document paint definition">ContextAttributes object</dfn>.
+ - A <dfn for="document paint definition">PaintRenderingContext2DSettings object</dfn>.
 
 Registering Custom Paint {#registering-custom-paint}
 ====================================================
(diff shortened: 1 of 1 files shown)


css-paint-api/Overview.bs
```diff
@@ -138,13 +138,22 @@ interface PaintWorkletGlobalScope : WorkletGlobalScope {
...
+</pre>
+
 <div class='note'>
     Note: The shape of the class should be:
     <pre class='lang-javascript'>
         class MyPaint {
             static get inputProperties() { return ['--foo']; }
             static get inputArguments() { return ['&lt;color>']; }
-            static get alpha() { return true; }
+            static get ContextAttributes() { return {alpha: true}; }
```
bfgeek: I think should be lowerCamelCase, "contextAttributes"

css-paint-api/Overview.bs
```diff
@@ -244,15 +253,15 @@ called, the user agent <em>must</em> run the following steps:
 
         2. <a for=list>Append</a> |parsedSyntax| to |inputArgumentSyntaxes|.
 
-    12. Let |alphaValue| be the result of <a>Get</a>(|paintCtor|, "alpha").
+    12. Let |alphaValue| be the result of <a>Get</a>(|paintCtor|, "ContextAttributes.alpha").
```
bfgeek: This should be...

12. Let |contextAttributesValue| be the result of <a>Get</a>(|paintCtor|, "contextAttributes").

13. Let |contextAttributes| be the result of <a>converting</a> |contextAttributesValue| to a {{PaintContextAttributes}}, .... etc etc etc

css-paint-api/Overview.bs
```diff
@@ -277,7 +286,7 @@ called, the user agent <em>must</em> run the following steps:
 
         - <a for="paint definition">input properties</a> being |inputProperties|.
 
-        - <a for="paint definition">context alpha flag</a> being |alpha|.
+        - <a for="paint definition">ContextAttributes object</a> being {alpha: |alpha|}.
```
bfgeek: ... being |contextAttributes|.

css-paint-api/Overview.bs
```diff
@@ -293,7 +302,7 @@ called, the user agent <em>must</em> run the following steps:
             - <a for="document paint definition">input argument syntaxes</a> being
                 |inputArgumentSyntaxes|.
 
-            - <a for="document paint definition">context alpha flag</a> being |alpha|.
+            - <a for="document paint definition">ContextAttributes object</a> being {alpha: |alpha|}.
```
bfgeek: as above.

css-paint-api/Overview.bs
```diff
@@ -138,13 +138,22 @@ interface PaintWorkletGlobalScope : WorkletGlobalScope {
 };
 </pre>
 
+The {{ContextAttributes}} allows the settings for the rendering context associated
+with the paint canvas. Right now, the {{ContextAttributes}} has a boolean flag
+indicating if the paint canvas contains an alpha channel.
+<pre class='idl'>
+dictionary ContextAttributes {
```
bfgeek: might be nice to namespace this for the spec e.g. {{PaintContextAttributes}}, but doesn't really matter as not user visible.

css-paint-api/Overview.bs
```diff
@@ -244,15 +253,15 @@ called, the user agent <em>must</em> run the following steps:
 
         2. <a for=list>Append</a> |parsedSyntax| to |inputArgumentSyntaxes|.
 
-    12. Let |alphaValue| be the result of <a>Get</a>(|paintCtor|, "alpha").
+    12. Let |contextAttributesValue| be the result of <a>Get</a>(|paintCtor|, "ContextAttributes").
```
bfgeek: "contextAttributes"

css-paint-api/Overview.bs
bfgeek: PaintContextAttributes to match below.

css-paint-api/Overview.bs
```diff
@@ -605,7 +614,7 @@ When the user agent wants to <dfn>invoke a paint callback</dfn> given |name|, |i
     8. Let |renderingContext| be the result of <a>create a PaintRenderingContext2D object</a> given:
```
bfgeek: has the "create a PaintRenderingContext2D object" algorithm been updated to receive a contextAttributes object?

css-paint-api/Overview.bs
bfgeek: @flackr thoughts?

flackr: It makes sense that we might eventually want to support settings like color space. However, in the event that we do this I think we might still only want to support a subset of CanvasRenderingContext2DSettings just like PaintRenderingContext2D implements a subset of CanvasRenderingContext2D. How about PaintRenderingContext2DSettings for consistency?

xidachen: Yes, I'd support that. PaintRenderingContext2DSettings is a better name, because it is more clear that it is used for PaintRenderingContext2D.
Question: do we need to start a discussion somewhere and bring other people onboard, like CSS discussion group?

css-paint-api/Overview.bs
xidachen: Good catch, I have updated that section.

xidachen:
@bfgeek @flackr The name of the dictionary is now "PaintRenderingContext2DSettings".

xidachen:
@bfgeek @flackr Gentle ping.

css-paint-api/Overview.bs
```diff
@@ -138,13 +138,22 @@ interface PaintWorkletGlobalScope : WorkletGlobalScope {
 };
 </pre>
 
+The {{PaintRenderingContext2DSettings}} allows the settings for the rendering context associated
```
flackr: This sentence is confusing, what do you mean "allows the settings"? Maybe contains or provides?

css-paint-api/Overview.bs
```diff
@@ -138,13 +138,22 @@ interface PaintWorkletGlobalScope : WorkletGlobalScope {
...
+with the paint canvas. Right now, the {{PaintRenderingContext2DSettings}} has a boolean flag
```
flackr: I think it would be more clear to have a bulleted list of what each value means. Also, perhaps instead of implying that we may add more settings we should say this provides a supported subset of the canvas context 2d settings (ideally with a link)

css-paint-api/Overview.bs
xidachen: Done

css-paint-api/Overview.bs
xidachen: Changed, please review the new commit.

xidachen:
Yay, thanks for reviewing it :)

xidachen:
I have changed the function name to contextOptions, please review.

PR #451, opened 2017-08-18
Add description for StylePropertyMapReadOnly::getAll

nainar:
@shans PTAL? Thanks!

### DIFF
--- a/css-typed-om/Overview.bs
+++ b/css-typed-om/Overview.bs
@@ -305,6 +305,29 @@ The <dfn method for=StylePropertyMap>append(DOMString <var>property</var>,
 </div>
 
 <div algorithm>
+    To <dfn>get all values from a StylePropertyMap</dfn>, run these steps:
+
+    1. If |property| does not start with two dashes (U+002D HYPHEN),
+        let |property| be |property| [=ASCII lowercased=].
+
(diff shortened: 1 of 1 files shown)


css-typed-om/Overview.bs
```diff
@@ -280,6 +280,29 @@ The <dfn method for=StylePropertyMap>append(DOMString <var>property</var>,
...
+        
+        1. If |property| is a [=list-valued property=],
+        return the list of |values| assigned to that property.
+        
+        2. If |property| is not a [=list-valued property=],
+        let |values| initially be the empty [=list=] and append 
+        the |value| assigned to the |property| to this list 
+        and return |values|.
+    
+    4. Else, return `null`.
```
shans: Not sure if this should be null or []. @tabatkins ? At any rate, good to merge.

tabatkins: Checking with @annevk and @domenic (who have also worked with developing the MultiMap pattern) to see if `getAll()` should return `undefined` or `[]` for a non-existent key.  (It definitely shouldn't be `null`, tho.)

annevk: Answer is [].

nainar:
Changed `null` to empty list. Merging change in now.

Issue #350, opened 2017-02-03
[all] Move Houdini APIs to `window.CSS`


tabatkins:
Per Seattle resolution (and TAG feedback), all the Houdini APIs should namespace themselves under the `CSS` object, rather than living on the global.

(Constructors still get put on the global by default, and that should be fine. We should add convenient aliases to `CSS` as appropriate, like `CSS.px(x)` for `CSSSimpleLength(x, 'px')`.)

dbaron:
#330 has some prior discussion (though specific to css-paint-api), including references to WebIDL work

shans:
For typed OM, this would just be Window.getComputedStyleMap.

Should that become CSS.getComputedStyleMap, or CSS.getComputedStyle?

bfgeek:
I've changed all the css-paint-api to CSS interface.

nainar:
With regards to @shans suggestion I vote for CSS.getComputedStyleMap to make it clear to developers if they misuse it

nainar:
As per #454 - removing css-typed-om label. Since the work pertaining to that spec seems to be done here.

majido:
@tabatkins  can we close this issue now? Is there any more work left to be done?

PR #454, opened 2017-08-22
[css-typed-om] #350 Rename styleMap and getComputedStyleMap

darrnshn:
(#350) Renames `Element.styleMap` to `Element.attributeStyleMap` and `Window.getComputedStyleMap` to `Element.computedStyleMap`. @shans  PTAL


shans:
Have a look at the resolution text in the linked issue. What we ended up deciding on is quite different to this.

darrnshn:
Cool, I've updated the patch. I removed the pseudo element parameter in `computedStyleMap` since it looks like it'll be handled by the pseudo element API or something? Also, there's a `CSSStyleRule.styleMap` which I didn't rename, should I?

PR #455, opened 2017-08-22
Add height/width serialization

nainar:
@shans PTAL? Thanks!


shans:
Does this deal with calc expressions?

tabatkins:
Ah, no it doesn't, and so I need to go revise the ones I just merged that do length/percentage. ^_^

(Check for <<length-percentage>> and serialize as that; the rules for it will need to output a calc().)

PR #456, opened 2017-08-22
Add serialization for the margin and margin-{bottom,left,right,top}

nainar:
@shans PTAL? Thanks!

### DIFF
--- a/css-typed-om/Overview.bs
+++ b/css-typed-om/Overview.bs
@@ -303,17 +303,6 @@ The <dfn method for=StylePropertyMap>append(DOMString <var>property</var>,
 </div>
 
 <div algorithm>
-    To <dfn>check if StylePropertyMap has a property</dfn>, run these steps:
-
-    1. Run the algorithm to [=get a value from a StylePropertyMap=] with property <var>property</var>,
-        1. If the algoritm returns a {{CSSStyleValue}} return true.
-
-        2. If the algorithm returns `null` return false.
-
-</div>
-
-
-<div algorithm>
     To <dfn>set a value on a StylePropertyMap</dfn>, run these steps:
 
     1. If |property| does not start with two dashes (U+002D HYPHEN),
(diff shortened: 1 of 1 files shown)


css-typed-om/Overview.bs
```diff
@@ -303,6 +303,17 @@ The <dfn method for=StylePropertyMap>append(DOMString <var>property</var>,
...
+
+    1. Run the algorithm to [=get a value from a StylePropertyMap=] with property <var>property</var>,
+        1. If the algoritm returns a {{CSSStyleValue}} return true.
+
+        2. If the algorithm returns `null` return false.
+
+</div>
+
+
+<div algorithm>
```
shans: what's this doing here?

css-typed-om/Overview.bs
```diff
@@ -2892,6 +2903,52 @@ depending on the property they came from:
         return "currentcolor".
     2. Otherwise, return the result of serializing the <<color>> value.
 
+: 'margin'
+::
+    1. Let |values| initially be the empty [=list=].
+    2. If 'margin-bottom' does not specify ''0'' for it's initial value,
```
shans: its

css-typed-om/Overview.bs
```diff
@@ -2892,6 +2903,52 @@ depending on the property they came from:
...
+        serialize 'margin-bottom' and append it to |values|.
+    3. If 'margin-left' does not specify ''0'' for it's initial value, 
+        serialize 'margin-left' and append it to |values|.
+    4. If 'margin-right' does not specify ''0'' for it's initial value, 
+        serialize 'margin-right' and append it to |values|.
+    5. If 'margin-top' does not specify ''0'' for it's initial value, 
+        serialize 'margin-top' and append it to |values|.
+    6. If 'margin' is set to the ''inherit'' keyword, append "inherit" to |values|.
+    7. Return the result of concatenating all the items in |values|,
+        separated by " " (U+0020 SPACE).
```
shans: Many of the 4-length properties have collapsed versions when all the lengths (or some of the lengths) are the same. Do we currently serialize collapsed versions? Do other browsers? If there's consensus on collapsed serialization then we should spec that.

nainar: Checked other browsers and fixed it to include collapsing. (Copied the Chrome implementation here)

css-typed-om/Overview.bs
```diff
@@ -2892,6 +2892,60 @@ depending on the property they came from:
...
+        1. |topValue| and |bottomValue| are not equivalent
+        2. Or |rightValue| and |leftValue| are not equivalent.
+    8. If |rightValue| and |leftValue| are not equivalent append |leftValue| to |values|.
+    9. If 'margin' is set to the ''inherit'' keyword, append "inherit" to |values|.
+    10. Return the result of concatenating all the items in |values|,
+        separated by " " (U+0020 SPACE).
+
+: 'margin-bottom'
+::
+    1. If the value specifies ''0'', exit the algorithm.
```
tabatkins: Exit and do what? When serializing 'margin-bottom', ''0px'' is still a valid value that needs to be serialized. The algorithm always needs to output a value.

(Also, in general, lengths need to be talked about in slightly more generic terms.  It's not about specifying ''0'', it's about the `<<length>>` value being equivalent to ''0px''.)

Same applies to all the 'margin-*' properties.

css-typed-om/Overview.bs
```diff
@@ -2892,6 +2892,60 @@ depending on the property they came from:
...
+    2. If the value is the ''inherit'' keyword,
+        return "inherit".
```
tabatkins: Hm, we should address this at a higher level. This doesn't address any of the other global keywords, and *every* property allows the global keywords. Go ahead and cut this line entirely; we need to intervene for these keywords before getting to any specific property.

css-typed-om/Overview.bs
```diff
@@ -2892,6 +2892,60 @@ depending on the property they came from:
...
+    3. If the value is of type <<margin-width>> return the
```
tabatkins: `<<margin-width>>` isn't a type.  It's either a `<<length>>` or `auto`; serialize these two branches.

css-typed-om/Overview.bs
```diff
@@ -2892,6 +2892,60 @@ depending on the property they came from:
...
+        consider 'margin-left' to be |leftValue|.
+    6. Append |rightValue| to |values| if one of the following conditions is true:
+        1. |rightValue| and |topValue| are not equivalent,
+        2. |topValue| and |bottomValue| are not equivalent
+        3. Or |rightValue| and |leftValue| are not equivalent.
+    7. Append |bottomValue| to |values| if one of the following conditions is true:
+        1. |topValue| and |bottomValue| are not equivalent
+        2. Or |rightValue| and |leftValue| are not equivalent.
+    8. If |rightValue| and |leftValue| are not equivalent append |leftValue| to |values|.
+    9. If 'margin' is set to the ''inherit'' keyword, append "inherit" to |values|.
```
tabatkins: Like the sub-properties, cut this line. We'll deal with the global keywords generically.

PR #457, opened 2017-08-22
[css-typed-om] Add serialization for 'opacity'

darrnshn:
Adds instructions to serialize the opacity property.

@shans PTAL?


shans:
LGTM

PR #458, opened 2017-08-22
Add text for serializing numbers/lengths

nainar:
@shans  take a look? Will send it to @tabatkins after your ok

### DIFF
--- a/css-typed-om/Overview.bs
+++ b/css-typed-om/Overview.bs
@@ -2845,11 +2845,33 @@ of serializations of the contained {{CSSTransformComponent}} objects.
 {{CSSURLImageValue}} objects are serialized to the string given by
 "url(" + {{CSSURLImageValue/url}} + ")".
 
-{{CSSFontFaceValue}} Serialization {#fontfacevalue-serialization}
+Units and Values Serialization {#unitsandvalues-serialization}
 -----------------------------------------------------------------
 
-{{CSSFontFaceValue}} objects are serialized to the value of their contained
-{{CSSFontFaceValue/fontFamilyName}}.
+This section covers how to serialize the various unit types in CSSOM
+
+: 'length'
+::
+    1. Let |values| initially be the empty [=list=].
+    2. If the <length> unit has a '-' sign character append it to |values|.
+    3. Append the result of serializing a 'number' to |values|. 
+    4. If the number is '0' skip this step. Else, append a unit identifier 
+        (a two-letter abbreviation) to |values|.
+    5. Return the result of concatenating all the items in |values|,
+        separated by " " (U+0020 SPACE).
+
+: 'number'
+::
+    1. Return the number respresented as a string.
+
+: 'percentage'
+::
+    // TODO
+
+: 'color'
+::
+    // TODO
+
 
 Serialization from CSSOM Values {#cssom-serialization}
 ------------------------------------------------------
--- a/css-typed-om/Overview.bs
+++ b/css-typed-om/Overview.bs
@@ -2845,32 +2845,49 @@ of serializations of the contained {{CSSTransformComponent}} objects.
 {{CSSURLImageValue}} objects are serialized to the string given by
 "url(" + {{CSSURLImageValue/url}} + ")".
 
-Units and Values Serialization {#unitsandvalues-serialization}
+{{CSSFontFaceValue}} Serialization {#fontfacevalue-serialization}
 -----------------------------------------------------------------
+ 
+{{CSSFontFaceValue}} objects are serialized to the value of their contained
+{{CSSFontFaceValue/fontFamilyName}}.
+
+Units and Values Serialization {#unitsandvalues-serialization}
+--------------------------------------------------------------
 
 This section covers how to serialize the various unit types in CSSOM
 
-: 'length'
-::
+<div algorithm="serialize a 'length' value">
+
+To <dfn>serialize a 'length' value</dfn>, run these steps:
     1. Let |values| initially be the empty [=list=].
-    2. If the <length> unit has a '-' sign character append it to |values|.
-    3. Append the result of serializing a 'number' to |values|. 
-    4. If the number is '0' skip this step. Else, append a unit identifier 
-        (a two-letter abbreviation) to |values|.
+    2. If the <length> value has a '-' sign character append it to |values|.
+    3. Append the result of [=serialize a <length> value=] to |values|. 
+    4. If the number is '0' skip this step. Else, append the <length> 
+        value's unit identifier (a two-letter abbreviation) to |values|.
     5. Return the result of concatenating all the items in |values|,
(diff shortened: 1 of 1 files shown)


css-typed-om/Overview.bs
```diff
@@ -2845,11 +2845,33 @@ of serializations of the contained {{CSSTransformComponent}} objects.
 {{CSSURLImageValue}} objects are serialized to the string given by
 "url(" + {{CSSURLImageValue/url}} + ")".
 
-{{CSSFontFaceValue}} Serialization {#fontfacevalue-serialization}
```
shans: Does this actually remove the FontFace serialization?

It's fine, we probably want to get rid of it, just strange to do both at once.

css-typed-om/Overview.bs
```diff
@@ -2845,11 +2845,33 @@ of serializations of the contained {{CSSTransformComponent}} objects.
...
 
-{{CSSFontFaceValue}} objects are serialized to the value of their contained
-{{CSSFontFaceValue/fontFamilyName}}.
+This section covers how to serialize the various unit types in CSSOM
+
+: 'length'
+::
+    1. Let |values| initially be the empty [=list=].
+    2. If the <length> unit has a '-' sign character append it to |values|.
+    3. Append the result of serializing a 'number' to |values|.
```
shans: Eventually we'll want to make sure that there's links between all these things

css-typed-om/Overview.bs
```diff
@@ -2845,11 +2845,33 @@ of serializations of the contained {{CSSTransformComponent}} objects.
...
-{{CSSFontFaceValue/fontFamilyName}}.
+This section covers how to serialize the various unit types in CSSOM
+
+: 'length'
+::
+    1. Let |values| initially be the empty [=list=].
+    2. If the <length> unit has a '-' sign character append it to |values|.
+    3. Append the result of serializing a 'number' to |values|. 
+    4. If the number is '0' skip this step. Else, append a unit identifier 
+        (a two-letter abbreviation) to |values|.
```
shans: which unit identifier?

css-typed-om/Overview.bs
nainar: Done. This was a typo

css-typed-om/Overview.bs
nainar: Done. Adding algorithm links

css-typed-om/Overview.bs
nainar: Mentioned that this needs to be the length's unit identifier.

css-typed-om/Overview.bs
```diff
@@ -2847,10 +2847,49 @@ of serializations of the contained {{CSSTransformComponent}} objects.
...
+Units and Values Serialization {#unitsandvalues-serialization}
+--------------------------------------------------------------
+
+This section covers how to serialize the various unit types in CSSOM
+
+<div algorithm="serialize a 'length' value">
+
+To <dfn>serialize a 'length' value</dfn>, run these steps:
+    1. Let |values| initially be the empty [=list=].
+    2. If the <length> value has a '-' sign character append it to |values|.
```
tabatkins: Lengths don't have characters; look for it being negative instead, and then set it to its absolute value.

css-typed-om/Overview.bs
```diff
@@ -2847,10 +2847,49 @@ of serializations of the contained {{CSSTransformComponent}} objects.
...
 {{CSSFontFaceValue/fontFamilyName}}.
 
+Units and Values Serialization {#unitsandvalues-serialization}
+--------------------------------------------------------------
+
+This section covers how to serialize the various unit types in CSSOM
+
+<div algorithm="serialize a 'length' value">
+
+To <dfn>serialize a 'length' value</dfn>, run these steps:
```
tabatkins: 'length' isn't a property. Same for all the others.

css-typed-om/Overview.bs
```diff
@@ -2847,10 +2847,49 @@ of serializations of the contained {{CSSTransformComponent}} objects.
...
+    1. Let |values| initially be the empty [=list=].
+    2. If the <length> value has a '-' sign character append it to |values|.
+    3. Append the result of [=serialize a <length> value=] to |values|.
```
tabatkins: What is this algorithm you're invoking? It looks like it's the same algorithm you're defining.  Maybe you meant to serialize the length's value as a number by invoking the "serialize a number" algorithm? (And if so, you want to keep the negative on there, as the number serializer will handle it for you.)

css-typed-om/Overview.bs
```diff
@@ -2847,10 +2847,49 @@ of serializations of the contained {{CSSTransformComponent}} objects.
...
+<div algorithm="serialize a 'length' value">
+
+To <dfn>serialize a 'length' value</dfn>, run these steps:
+    1. Let |values| initially be the empty [=list=].
+    2. If the <length> value has a '-' sign character append it to |values|.
+    3. Append the result of [=serialize a <length> value=] to |values|. 
+    4. If the number is '0' skip this step. Else, append the <length> 
+        value's unit identifier (a two-letter abbreviation) to |values|.
+    5. Return the result of concatenating all the items in |values|,
+        separated by " " (U+0020 SPACE).
```
tabatkins: You don't want space separation here. Just build a string directly by concatenation in each step; no need to indirect thru a list.

css-typed-om/Overview.bs
```diff
@@ -2847,10 +2847,49 @@ of serializations of the contained {{CSSTransformComponent}} objects.
...
+</div>
+
+<div algorithm="serialize a 'number' value">
+
+To <dfn>serialize a 'number' value</dfn>, run these steps:
+    1. Return the number respresented as a string.
```
tabatkins: If we're trying to replace CSSOM for this, we need to go into the same level of detail as it does, particularly regarding the minimum required amount of decimal precision, and if we ever output scinot.

css-typed-om/Overview.bs
```diff
@@ -2847,10 +2847,49 @@ of serializations of the contained {{CSSTransformComponent}} objects.
...
+</div>
+
+
+<div algorithm="serialize a 'percentage' value">
+
+To <dfn>serialize a 'percentage' value</dfn>, run these steps:
+    // TODO
```
tabatkins: Should just be "serialize a number" and then append a "%" to the end of the result, right?

css-typed-om/Overview.bs
```diff
@@ -2847,10 +2847,49 @@ of serializations of the contained {{CSSTransformComponent}} objects.
...
+</div>
+
+
+<div algorithm="serialize a 'color' value">
+
+To <dfn>serialize a 'color' value</dfn>, run these steps:
+    // TODO
```
tabatkins: This one's gonna be quite complicated; you're not gonna want to have it in the same PR.

PR #459, opened 2017-08-22
Add serialization for stroke geometry

nainar:
@shans TAL? This should be easy?

### DIFF
--- a/css-typed-om/Overview.bs
+++ b/css-typed-om/Overview.bs
@@ -2925,6 +2925,18 @@ depending on the property they came from:
         return the result of serializing the <<length>> value.
     3. Otherwise, return the result of serializing the <<percentage>> value.
 
+: 'stroke-align'
+::
+    1. Return the result of serializing the keyword value.
+
+: 'stroke-break'
+::
+    1. Return the result of serializing the keyword value.
+
+: 'stroke-linecap'
+::
+    1. Return the result of serializing the keyword value.
+
 : 'top'
 ::
     1. If the value is the ''auto'' keyword,


css-typed-om/Overview.bs
```diff
@@ -2903,6 +2903,52 @@ depending on the property they came from:
         return "currentcolor".
     2. Otherwise, return the result of serializing the <<color>> value.
 
+: 'stroke-align'
+::
+    1. If 'stroke-align' is one of the following keywords,
```
shans: what if it isn't?

nainar: Removed the if conditions where no other course of action is possible. TAL?

css-typed-om/Overview.bs
```diff
@@ -2903,6 +2903,49 @@ depending on the property they came from:
         return "currentcolor".
     2. Otherwise, return the result of serializing the <<color>> value.
 
+: 'stroke-align'
+::
+    1. Return the string representation of the keyword:
+        * center,
+        * inset,
+        * outset.
```
shans: I don't know that you need to explicitly list the keywords even. Maybe get @tabatkins to take a look.

tabatkins: Yeah, no need to list these explicitly. Just say "return the result of serializing the keyword value".

css-typed-om/Overview.bs
```diff
@@ -2903,6 +2903,49 @@ depending on the property they came from:
...
+
+: 'stroke-linejoin'
+::
+    1. Let |values| initially be the empty [=list=].
+    2. If 'stroke-linejoin' has one of the following keywords,
+        append the string representation of the keyword to |values|:
+        * crop,
+        * arcs,
+        * miter.
+    3. If 'stroke-linejoin' has one of the following keywords,
```
tabatkins: 'stroke-linecap'

css-typed-om/Overview.bs
```diff
@@ -2903,6 +2903,49 @@ depending on the property they came from:
...
+::
+    1. Return the string representation of the keyword:
+        * butt,
+        * round,
+        * square.
+
+: 'stroke-linejoin'
+::
+    1. Let |values| initially be the empty [=list=].
+    2. If 'stroke-linejoin' has one of the following keywords,
```
tabatkins: Instead list the keyword that it can't be (the one that it behaves as if omitted). This avoids us having to update this in the future if we add more keywords.  Same for step 3.

css-typed-om/Overview.bs
```diff
@@ -2903,6 +2903,49 @@ depending on the property they came from:
...
+        append the string representation of the keyword to |values|:
+        * crop,
+        * arcs,
+        * miter.
+    3. If 'stroke-linejoin' has one of the following keywords,
+        append the string representation of the keyword to |values|:
+        * bevel,
+        * round,
+        * stupid.
+    4. Return the result of concatenating all the items in |values|,
```
tabatkins: It might be empty, need to handle that case too (emit the shortest value that triggers the initial-value behavior; check browsers to see what they do).

css-typed-om/Overview.bs

PR #460, opened 2017-08-22
Add serialization for left/right/top/bottom

darrnshn:
Add serialization procedures of left/right/top/bottom.

@shans PTAL?


shans:
LGTM, @tabatkins ?

PR #461, opened 2017-08-22
Serialize property display

jm318:
Add description for serializing 'display' property.

### DIFF
--- a/css-typed-om/Overview.bs
+++ b/css-typed-om/Overview.bs
@@ -300,7 +300,7 @@ The <dfn method for=StylePropertyMap>append(DOMString <var>property</var>,
 
     3. If {{StylePropertyMap}}’s [=property model=] contains an entry for |property|,
         return the first value found in that entry.
-    
+
     4. Else, return `null`.
 </div>
 
@@ -2275,7 +2275,7 @@ objects, the {{CSSURLImageValue/url}} attribute contains the URL that references
     when called,
     perform the following steps:
 
-    1. If the |url| passed into the constructor doesn't correctly 
+    1. If the |url| passed into the constructor doesn't correctly
         parse as a <<url>>, throw a {{TypeError}} and exit this algorithm.
     2. Else, return a new {{CSSURLImageValue}}
         with its {{CSSURLImageValue/url}} internal slot
@@ -2879,7 +2879,7 @@ depending on the property they came from:
     3. If 'border-image-slice' does not specify ''100%'' for all sides and omits the ''border-image-slice/fill'' keyword,
         serialize 'border-image-slice' and append it to |values|.
     4. If 'border-image-width' does not specify ''1'' for all sides,
-        append "/ " (U+002F FORWARD SLASH followed by U+0020 SPACE) 
+        append "/ " (U+002F FORWARD SLASH followed by U+0020 SPACE)
         to the result of serializing 'border-image-width' and append it to |values|.
     5. If 'border-image-outset' does not specify ''0'' for all sides:
         1. If the previous 'border-image-width' step did not append anything to |values|,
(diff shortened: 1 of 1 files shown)


jm318:
@shans PTAL. Thanks!

jm318:
@shans PTAL. Thanks!

shans:
LGTM. @tabatkins ?

css-typed-om/Overview.bs
```diff
@@ -2903,6 +2903,44 @@ depending on the property they came from:
         return "currentcolor".
     2. Otherwise, return the result of serializing the <<color>> value.
 
+: 'display'
+::
+    1. Let |values| initially be the empty [=list=].
+    2. If 'display-outside' or 'display-inside' is specified:
```
tabatkins: These properties don't exist. You probably mean the `<<display-inside>>` and `<<display-outside>>` values. ^_^

css-typed-om/Overview.bs
```diff
@@ -2903,6 +2903,44 @@ depending on the property they came from:
...
+        1. Append value of 'display-outside' to |values|.
```
tabatkins: The value is an abstract CSS value; need to serialize it to produce a JS value. (Same applies to all the steps.)

css-typed-om/Overview.bs
```diff
@@ -2903,6 +2903,44 @@ depending on the property they came from:
...
+        2. Append value of 'display-inside' to |values|.
+        3. Goto step 7.
```
tabatkins: This means you won't hit the list-item step at all if you specified something like `display: block list-item`.

css-typed-om/Overview.bs
```diff
@@ -2903,6 +2903,44 @@ depending on the property they came from:
...
+    3. If 'display-listitem' is specified:
+        1. Append ''list-item'' to |values|.
```
tabatkins: Append a string, `"list-item"`, not a CSS value `''list-item''`.

css-typed-om/Overview.bs
```diff
@@ -2903,6 +2903,44 @@ depending on the property they came from:
...
+        1. Append its value to |values|.
+        2. Goto step 7.
+    5. If 'display-box' is specified:
+        1. Append its value to |values|.
+        2. Goto step 7.
+    6. If 'display-legacy' is specified:
+        1. Append its value to |values|.
+        2. Goto step 7.
+    7. If |values| is:
+        1. ''block flow'', return 'block'.
```
tabatkins: |values| is a list of strings, not a sequence of CSS values. Needs to be `«"block", "flow"»`, per [the Infra Standard](https://infra.spec.whatwg.org/#lists).

tabatkins: And `'block'` isn't a valid property, and should be a string instead.

css-typed-om/Overview.bs
```diff
@@ -2903,6 +2903,44 @@ depending on the property they came from:
...
+        5. ''run-in flow'', return ''run-in''.
+        6. ''list-item block flow'', return ''list-item''.
+        7. ''list-item inline flow'', return ''inline list-item''.
+        8. ''block flex'', return ''flex''.
+        9. ''inline flex'', return ''inline-flex''.
+        10. ''block grid'', return ''grid''.
+        11. ''inline ruby'', return ''ruby''.
+        12. ''block table'', return ''table''.
+        13. ''inline table'', return ''inline-table''.
+        14. [=list/empty=], return "none".
```
tabatkins: How does it end up empty, and how does emptiness equate to `display: none`? `none` is handled by the `<<display-box>>` clause in step 5, right?

css-typed-om/Overview.bs
jm318: Done, thanks for pointing out! :)

css-typed-om/Overview.bs
jm318: Done, thanks!

css-typed-om/Overview.bs
jm318: Yes, and I understand this is what the spec says according to "|", i.e. these are mutually exclusive?

css-typed-om/Overview.bs
jm318: Done

css-typed-om/Overview.bs
jm318: Done, thanks for pointing out!

css-typed-om/Overview.bs
jm318: I thought it would be empty if property is simply "display:", but I guess it's illegal anyway. So I removed the case for empty-list.

jm318:
tabatkins@ I've addressed your comments, PTAL, thanks!

PR #462, opened 2017-08-22
Add serialization for box-sizing

nainar:
@shans  PTAL?

### DIFF
--- a/css-typed-om/Overview.bs
+++ b/css-typed-om/Overview.bs
@@ -2895,6 +2895,10 @@ depending on the property they came from:
     8. Return the result of concatenating all the items in |values|,
         separated by " " (U+0020 SPACE).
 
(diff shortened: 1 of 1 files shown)


css-typed-om/Overview.bs
```diff
@@ -2897,6 +2897,13 @@ depending on the property they came from:
     8. Return the result of concatenating all the items in |values|,
         separated by " " (U+0020 SPACE).
 
+: 'box-sizing'
+::
+    1. Return the string representation of the keyword:
```
tabatkins: Just serialize the value as a keyword. No need to specify the keywords; it just means we'd have to edit this when new keywords are added.

nainar:
@tabatkins I have made the changes you asked for, sorry for the late reply on this. Could you TAL?

PR #463, opened 2017-08-22
Serialization description for position property

jm318:
@shans PTAL, thanks!


css-typed-om/Overview.bs
```diff
@@ -2903,6 +2903,9 @@ depending on the property they came from:
         return "currentcolor".
     2. Otherwise, return the result of serializing the <<color>> value.
 
+: 'position'
+::
+    1. Return value of 'position'.
```
tabatkins: The "value of 'position'" is a browser-internal abstract CSS value, which can't be returned directly to JS.  It needs to be serialized as a keyword, like in the other examples.

jm318: Done, thanks!

jm318:
@tabatkins, thanks for the review! PTAL.

PR #464, opened 2017-08-24
Use [Exposed] consistently (needs fixup, replace "???")

zcorpan:
Needed to fix https://github.com/heycam/webidl/issues/365.

---

I wasn't sure what to put in `Exposed` in some cases, please fix those as appropriate.

cc @annevk @tobie

### DIFF
--- a/worklets/Overview.bs
+++ b/worklets/Overview.bs
@@ -288,6 +288,7 @@ The {{Worklet}} object provides the capability to add module scripts into its as
 {{WorkletGlobalScope}}s and invoke their methods.
 
 <pre class='idl'>
+[Exposed=???]
 interface Worklet {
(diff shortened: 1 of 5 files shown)


worklets/Overview.bs
```diff
@@ -288,6 +288,7 @@ The {{Worklet}} object provides the capability to add module scripts into its as
 {{WorkletGlobalScope}}s and invoke their methods.
 
 <pre class='idl'>
+[Exposed=???]
```
annevk: Window I suspect?

saschanaz:
I think it's now safe to close this.

zcorpan:
Yes, looks like all Houdini specs now use `[Exposed]`

Issue #471, opened 2017-09-24
How are paint worklets picked?


dglazkov:
From https://groups.google.com/a/chromium.org/d/msg/blink-dev/Jex3idOld48/XWkgUzlxBQAJ

@bzbarsky:

> What the spec defines is that a UA must have at least two paint worklets
> and assign work to them somehow.  The spec does not require randomness
> of ordering of work assignment, randomness in which worklet is chosen,
> or anything else.
> 
> This unfortunately makes it fairly easy to create worklets that depend
> on specific assignment algorithms, and I expect pages to end up doing
> just that if browsers do anything deterministic here.  At which point
> we'll end up having to specify those deterministic algorithms, and
> whatever their limitations are (e.g. only two paint worklets, period).

PR #472, opened 2017-09-25
[worklets] Add language for non-derminism in a note.

bfgeek:
This adds a note that user-agents must introduce some non-determinism into the selection process for a paint worklet global scope.

This came up in blinks intent-to-ship thread:
https://groups.google.com/a/chromium.org/forum/#!topic/blink-dev/Jex3idOld48

Issue: #471


foolip:
Some thoughts on how to test this. Could the spec have a limit to say that the same scope cannot be used more than n times in a row, where n is a large number? Then the implementation could be to randomly pick a scope and randomly switch, but a test that checks that it does eventually switch would both be supported by what the spec says, and would pass in a reasonable time frame.

PR #476, opened 2017-10-13
CSSNumericValue.from() -> CSSNumericValue.to()

nainar:
There were still 2 references to CSSNumericValue.from() -> I changed those to CSSNumericValue.to()


tabatkins:
`from` wasn't turned into `to`; `from` was the more generic factory function. It was replaced by `parse` in the first instance, and the CSSUnitValue constructor in the second instance (or `CSS.px()`).

nainar:
Closing in favor of #480

Issue #310, opened 2016-09-22
[css-typed-om]  Consider using properties in addition to .get/.set


esprehn:
Real properties are faster since we can cache the property lookup, ex.

```js
element.styleMap.color // => JS VM can cache the lookup for color
```

It also lets the getter/setter cache the "color" => engine representation conversion where most (all?) engines convert to an enum as there's a specific method for every property.

Note we'd still need .get(name) and .set(name) to deal with custom properties, since we don't want to add more named getters, ex. `map.get("--foo").` vs `map["--foo"]`.

The disadvantage of this approach is that it means StyleMap has 400+ properties like CSSStyleDeclaration does, and every new CSS property is implicitly adding new properties. It also means if maps get a new method called "filter", and we wanted StyleMap to be really maplike, we'd end up colliding with CSS's filter property.

@bzbarsky

bzbarsky:
The name collision thing can be addressed by having an extra object:

    element.styleMap.makeUpNameHere.color

That doesn't do anything for the other disadvantages, though.

shans:
Another disadvantage is that people might think
```element.styleMap.color.red += 20;```
should work.

PR #481, opened 2017-10-16
[css-typed-om] Remove issue #310 from text.

darrnshn:
Issue #310 was resolved to be moved to css-typed-om 2, so we can remove it from the L1 spec.

@shans PTAL?


nainar:
add it to Typed OM 2 instead?

darrnshn:
done, PTAL again?

PR #484, opened 2017-10-17
Add clarification about custom properties in StylePropertyMaps

darrnshn:
#276 

Not sure if the wording is enough for declared and inline StylePropertyMap. Also not sure how to do the wording for the computed StylePropertyMap. 

@shans PTAL?


tabatkins:
Handled in 2519d4b and a preceding commit about computed style.

Issue #492, opened 2017-10-19
[css-typed-om] What is CSSMathValue.type?


darrnshn:
I'm probably missing something here, but what is [CSSMathValue.type](https://drafts.css-houdini.org/css-typed-om-1/#dom-cssmathvalue-type)? The spec mentions what type of a CSSMathValue should be, but that type links to [CSSNumericValue.type](https://drafts.css-houdini.org/css-typed-om-1/#cssnumericvalue-type), which is a map not a string.

tabatkins:
It's a remnant of the world before we made things more complex. This seems like basically a dupe of #482, right?

darrnshn:
Ah right, does that mean we could just remove CSSMathValue.type, since it inherits CSSNumericValue's type() method?

tabatkins:
Yeah, we should.

nainar:
Fixed by #497

PR #497, opened 2017-10-25
Remove CSSMathValue.type

nainar:
This addresses issue #492.

### DIFF
--- a/css-typed-om/Overview.bs
+++ b/css-typed-om/Overview.bs
@@ -1412,7 +1412,6 @@ are represented as {{CSSUnitValue}}s.
     interface CSSUnitValue : CSSNumericValue {
         attribute double value;
         attribute DOMString unit;
-        readonly attribute DOMString type;
     };
 </xmp>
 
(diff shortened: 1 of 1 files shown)


css-typed-om/Overview.bs
```diff
@@ -1654,29 +1653,6 @@ of all the "math" operations.
...
-        : {{CSSMathProduct}}
-        :: The [=type=] is the result of [=multiplying=] the [=types=]
-            of each of the [=list/items=] in its {{CSSMathProduct/values}} internal slot.
-
-        : {{CSSMathNegate}}
-        : {{CSSMathInvert}}
-        :: The [=type=] is the same as the [=type=] of its {{CSSMathNegate/value}} internal slot,
-            but with all [=map/values=] negated.
-    </dl>
-</div>
```
tabatkins: No, all of this needs to stay - it defines the type of a CSSMathValue. We just need to remove the `.type` member in the IDL, since that's carried by the superclass.  The superclass's definition of `.type` will then figure out how to translate the type into some JS object.

darrnshn: I think there's actually two we need to remove: CSSMathValue.type and CSSUnitValue.type?

tabatkins: Yeah.

nainar: Ahhh I for some reason thought this had been copied twice. :/ Gotcha. Making the change. g

Issue #479, opened 2017-10-14
[css-paint-api] Specify explicitly that PaintSize.{width, height} may be fractional


chrishtr:
The paint API specifies that the width and height fields of PaintInfo are doubles, but
elsewhere in the spec (https://drafts.css-houdini.org/css-paint-api/#2d-rendering-context)
says to "Set bitmap dimensions for the context’s output bitmap to width and height", which
implies that the width and height are integers.

Instead, I suggest that phrase should be changed to "Set bitmap dimensions for the context’s output bitmap to the rounded values of width and height".

(Note: the actual device-pixel backing buffer may have a different size than the output bitmap. This
is accounted for by devicePixelRatio.)

chrishtr:
Also, an un-rounded width and height plus devicePixelRatio and specifying the top-left
of the painting canvas to be positioned at the device pixel location of the top-left of where
the element will actually be drawn to the screen, allows the developer, if they wish, to do pixel-perfect painting.

This is why it's important for width and height to be doubles.

bfgeek:
This has been fixed by PR #500

PR #500, opened 2017-10-26
[PaintWorklet] Explicit state the PaintSize could be fractional

xidachen:
It was pointed out here: https://github.com/w3c/css-houdini-drafts/issues/479, that the spec says to "Set bitmap dimensions for the context’s output bitmap to width and height", which implies that the width and height are integers.

This pull request makes change to say rounded values of width and height, which makes it explicit.


xidachen:
Wow, that's fast. Thank you :)

Issue #508, opened 2017-11-07
[css-paint-api] Calls to paint function should provide the snapped concrete size.


flackr:
[Section 8: Drawing an image](https://drafts.css-houdini.org/css-paint-api-1/#drawing-an-image) specifies that the Concrete Object Size is used when invoking the paint callback, however in the case of subpixel sizes the actual paint size depends on the browser's subpixel snapping.

For [example](http://jsbin.com/xoteqif/edit?html,js,output), given a device with a scale factor of 1.25 if drawing a series of boxes of size 9px, the physical screen size is 11.25px. This means that one in every 4 boxes will be drawn a physical pixel taller than the rest.

There are 4 options I can think of although I believe we should avoid the first two as neither allows developers to do the right thing:
1) Don't change spec, causing all paintings to be blurry (scaled up or down to fit pixel snapping).
2) Don't change spec, draw at subpixel position causing paintings to be blurry.
3) Adjust concrete object size to be snapped size / devicePixelRatio (e.g. in this example pass in either 11 / 1.25 = 8.8px or 12 / 1.25 = 9.6px). Edges of the drawn object size continue to be sharp and developers scaling by devicePixelRatio and accounting for the object size get pixel perfect drawings.
4) Add backingStorePixelRatio(X|Y) which maps concrete size to physical pixel size.

My preference is for option 3 given it is a small change, and it produces crisp edges for developers naively painting to their size, and do the right thing for developers taking into account the devicePixelRatio as well.

flackr:
As en example of each of these strategies, see this [code example](http://output.jsbin.com/wenutuh) and blown up image below:
![paintworklet-subpixel-examples](https://user-images.githubusercontent.com/366761/32522018-76d6492e-c3ca-11e7-93f1-3b68fdb1f51f.png)

PR #518, opened 2017-11-14
[CSS Paint API] Pass snapped concrete size to paint function

xidachen:
This pull request address the issue here:
https://github.com/w3c/css-houdini-drafts/issues/508

Basically, instead of passing the concrete object size to the paint function, we let it pass the round(concrete object size) / device pixel ratio.

Please review. @bfgeek @flackr

### DIFF
--- a/css-paint-api/Overview.bs
+++ b/css-paint-api/Overview.bs
@@ -452,8 +452,8 @@ Note: The user agent can optionally defer drawing images which are <a>paint-inva
 
 The <a>draw a paint image</a> function is invoked by the user agent during the <a>object size
 negotiation</a> algorithm which is responsible for rendering an <<image>>, with |snappedConcreteObjectSize|
-defined as follows. Let |concreateObjectSize| be the <a>concrete object size</a> of the <a>box</a>.
-The |snappedConcreateObjectSize| is usually the same as the |concreateObjectSize|. However, the
+defined as follows. Let |concreteObjectSize| be the <a>concrete object size</a> of the <a>box</a>.
+The |snappedConcreateObjectSize| is usually the same as the |concreteObjectSize|. However, the
 user agent may adjust the size such that it paints to pixel boundaries. If it does, the user agent
 should adjust the |snappedConcreateObjectSize| by the proportional change from its original size
 such that the <<paint()>> function can adjust the drawing accordingly.
--- a/css-paint-api/Overview.bs
+++ b/css-paint-api/Overview.bs
@@ -453,9 +453,9 @@ Note: The user agent can optionally defer drawing images which are <a>paint-inva
 The <a>draw a paint image</a> function is invoked by the user agent during the <a>object size
 negotiation</a> algorithm which is responsible for rendering an <<image>>, with |snappedConcreteObjectSize|
 defined as follows. Let |concreteObjectSize| be the <a>concrete object size</a> of the <a>box</a>.
-The |snappedConcreateObjectSize| is usually the same as the |concreteObjectSize|. However, the
+The |snappedConcreteObjectSize| is usually the same as the |concreteObjectSize|. However, the
 user agent may adjust the size such that it paints to pixel boundaries. If it does, the user agent
-should adjust the |snappedConcreateObjectSize| by the proportional change from its original size
+should adjust the |snappedConcreteObjectSize| by the proportional change from its original size
 such that the <<paint()>> function can adjust the drawing accordingly.
 
 For the purposes of the <a>object size negotiation</a> algorithm, the paint image has no
--- a/css-paint-api/Overview.bs
+++ b/css-paint-api/Overview.bs
@@ -452,11 +452,11 @@ Note: The user agent can optionally defer drawing images which are <a>paint-inva
 
 The <a>draw a paint image</a> function is invoked by the user agent during the <a>object size
 negotiation</a> algorithm which is responsible for rendering an <<image>>, with |snappedConcreteObjectSize|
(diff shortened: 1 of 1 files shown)


css-paint-api/Overview.bs
```diff
@@ -630,7 +631,7 @@ When the user agent wants to <dfn>invoke a paint callback</dfn> given |name|, |i
         <a>computed value</a>'s for properties listed in |inputProperties|.
 
     8. Let |renderingContext| be the result of <a>create a PaintRenderingContext2D object</a> given:
-            - "width" - The width given by |concreteObjectSize|.
+            - "width" - The  width given by |concreteObjectSize|.
```
bfgeek: new space?

css-paint-api/Overview.bs
```diff
@@ -451,8 +451,9 @@ Note: The user agent can optionally defer drawing images which are <a>paint-inva
 </div>
 
 The <a>draw a paint image</a> function is invoked by the user agent during the <a>object size
-negotiation</a> algorithm which is responsible for rendering an <<image>>, with |concreteObjectSize|
-set to the <a>concrete object size</a> of the <a>box</a>.
+negotiation</a> algorithm which is responsible for rendering an <<image>>, with |snappedConcreteObjectSize|
+set to the rounded <a>concrete object size</a> of the <a>box</a>, divided by the
```
flackr: I don't think this phrasing captures pixel snapping correctly. Also, AFAICT, concrete object size is in CSS pixels. Rounding the concrete object size would suggests that
* this only affects objects whose size is non-integral - this is not true as it affects integral sizes when they are rastered into a non-integral transformed space (i.e. non-integer devicePixelRatio).
* we always snap an object of a particular size in the same direction, but it actually depends on the object's location.
* all browsers snap the same way. This seems to be the case but should probably not be required.
Instead, I think we should specify this in a way that allows the browser to use its own subpixel snapping determination.

xidachen:
flackr@: I changed the description of |snappedConcreteSize|, let me know if it is better or worse.

css-paint-api/Overview.bs
```diff
@@ -451,8 +451,12 @@ Note: The user agent can optionally defer drawing images which are <a>paint-inva
 </div>
 
 The <a>draw a paint image</a> function is invoked by the user agent during the <a>object size
-negotiation</a> algorithm which is responsible for rendering an <<image>>, with |concreteObjectSize|
-set to the <a>concrete object size</a> of the <a>box</a>.
+negotiation</a> algorithm which is responsible for rendering an <<image>>, with |snappedConcreteObjectSize|
+defined as follows. Let |concreateObjectSize| be the <a>concrete object size</a> of the <a>box</a>, and
+|deviceObjectSize| be the |concreateObjectSize| multiplied by the {{PaintWorkletGlobalScope/devicePixelRatio}}.
+If the |deviceObjectSize| is non-integral, then adjust it to align with pixel boundary. Finally, set
```
flackr: It's not a requirement to snap, but the browser is allowed to. How about something like this?

The |snappedConcreteObjectSize| is usually the same as the |concreteObjectSize|; however, the user agent may adjust the size such that it paints to pixel boundaries. If it does, the user agent should adjust the |snappedConcreteObjectSize| by the proportional change from its original size so that the paint function can adjust the drawing accordingly.

xidachen: Done.

css-paint-api/Overview.bs
```diff
@@ -452,11 +452,11 @@ Note: The user agent can optionally defer drawing images which are <a>paint-inva
 
 The <a>draw a paint image</a> function is invoked by the user agent during the <a>object size
 negotiation</a> algorithm which is responsible for rendering an <<image>>, with |snappedConcreteObjectSize|
-defined as follows. Let |concreateObjectSize| be the <a>concrete object size</a> of the <a>box</a>, and
-|deviceObjectSize| be the |concreateObjectSize| multiplied by the {{PaintWorkletGlobalScope/devicePixelRatio}}.
-If the |deviceObjectSize| is non-integral, then adjust it to align with pixel boundary. Finally, set
-|snappedConcreateObjectSize| to the adjusted |deviceObjectSize| divided by the
-{{PaintWorkletGlobalScope/devicePixelRatio}}.
+defined as follows. Let |concreateObjectSize| be the <a>concrete object size</a> of the <a>box</a>.
```
flackr: s/concreateObjectSize/concreteObjectSize/ here and below.

xidachen: Fixed :)

css-paint-api/Overview.bs
```diff
@@ -452,8 +452,8 @@ Note: The user agent can optionally defer drawing images which are <a>paint-inva
 
 The <a>draw a paint image</a> function is invoked by the user agent during the <a>object size
 negotiation</a> algorithm which is responsible for rendering an <<image>>, with |snappedConcreteObjectSize|
-defined as follows. Let |concreateObjectSize| be the <a>concrete object size</a> of the <a>box</a>.
-The |snappedConcreateObjectSize| is usually the same as the |concreateObjectSize|. However, the
+defined as follows. Let |concreteObjectSize| be the <a>concrete object size</a> of the <a>box</a>.
+The |snappedConcreateObjectSize| is usually the same as the |concreteObjectSize|. However, the
```
flackr: s/snappedConcreateObjectSize/snappedConcreteObjectSize here and below

xidachen: Done.

Issue #498, opened 2017-10-26
[css-typed-om] Should CSSNumericValue attributes accept doubles?


darrnshn:
CSSMathNegate has a [CSSNumericValue 'value' attribute](https://drafts.css-houdini.org/css-typed-om-1/#cssmathnegate), so you can't set it to a double. But the spec also says: "Any place that accepts a CSSNumericValue also accepts a raw double". So should you be able to set CSSMathNegate.value to a double, or does 'place' mean like a function?

darrnshn:
On a similar note, the [various transform style](https://drafts.css-houdini.org/css-typed-om-1/#transformvalue-objects) constructors seem to be inconsistent about whether they're taking CSSNumericValues (e.g. translate) or CSSNumberishes (e.g. rotate). Is there a reason for this?

tabatkins:
Ah, that attribute should be a `CSSNumberish`, thanks.

darrnshn:
Cool thanks! In general, when should something be a `CSSNumericValue` vs `CSSNumberish`? Based on the "Any place that accepts a CSSNumericValue also accepts a raw double", I'm guessing the only place where you can have a `CSSNumericValue` is readonly attributes (since they don't "accept" `CSSNumericValues`) and everything other occurrence of `CSSNumericValue` should be a `CSSNumberish`?

tabatkins:
Any input that can potentially accept a CSS `<number>` (or, in TypedOM language, any place where a CSSUnitValue with a type of "number" would be valid) should be a `CSSNumberish`.  Many numeric attributes don't qualify; the difference in the transform style constructors is deliberate, as translate() only takes lengths.

Output should always be `CSSNumericValue`, as we'll always output TypedOM objects.  (So yeah, readonly attributes are `CSSNumericValue`, read/write attributes are `CSSNumberish` *if* they can accept a `<number>` value.)

(`CSSKeywordish` is more absolute - *all* keyword-accepting input should be `CSSKeywordish`. There's no conglomeration of value spaces to worry about here.)

-------

Note: I wrote my earlier comment based on the first comment in the thread, coming straight off of the email. I hadn't seen the second comment. I was specifically referring to `CSSMathNegate` needing to be switched to `CSSNumberish`, not referring to the transform stuff.

Issue #513, opened 2017-11-09
[css-typed-om] What to do when CSSNumericValues form a cycle?


darrnshn:
With mutable CSSNumericValues, I think it's possible to form cycles:
```javascript
let a = new CSSMathNegate(0) // a = (- 0)
let b = new CSSMathNegate(a) // b = (- a)
a.value = b // b = (- a) and a = (- b)
```
If this is possible, should `a.value = b` throw an error? Otherwise, type checking will get kinda weird and serialization will be impossible.

**Follow up question**: If we disallow cycles, we can still form a directed acyclic graph. I'm guessing that's allowed, with type checking propagating through the graph when a node changes their `value`? (i.e. a node has to notify its 'parents' when its type changes)

tabatkins:
Heh, don't even need to invoke a second object:

```
let a = new CSSMathNegate(0);
a.value = a; // ?????
```

I think the fix for this is, upon `a.value = b` (or equivalent construction), crawl b's value and see if the a object lives anywhere in it.

I think this still allows, say, `new CSSMathSum(a, a, a, a, a)` to work just fine.

Your follow-up question is more difficult, and means the resolution we made today for CSSUnitValues is kinda useless...

tabatkins:
Actually, if we solve https://github.com/w3c/css-houdini-drafts/issues/494 in the way I want (making `.value`/`.values` readonly), this goes away, as you can't construct cycles in the first place.

darrnshn:
This is pretty much the same as #494. Gonna close this.

PR #526, opened 2017-12-04
Change CSSMathNegate and CSSMathInvert to take numberishes

darrnshn:
Issue #498 .

@nainar PTAL

### DIFF
--- a/css-typed-om/Overview.bs
+++ b/css-typed-om/Overview.bs
@@ -1532,12 +1532,12 @@ interface CSSMathProduct : CSSMathValue {
 
 [Constructor(CSSNumberish arg)]
 interface CSSMathNegate : CSSMathValue {
-    attribute CSSNumericValue value;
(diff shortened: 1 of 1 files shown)


nainar:
lgtm on my end. @tabatkins

tabatkins:
This needs additional prose saying that the values, on setting, must rectify a numberish value. (The stored value should always be a CSSNumericValue, but you're allowed to set it with a plain JS number; it just turns into a `CSS.number(x)` automatically for you.)

darrnshn:
Hmm, with the resolution of #513, all the attributes of `CSSMathValues` will become readonly. Would it make more sense then to change all the `CSSNumberish` attributes to `CSSNumericValue`, since we can no longer set the value to a raw number?

tabatkins:
Yeah, they're all CSSNumericValue currently, so I'll go ahead and close this.

PR #544, opened 2017-12-13
StylePropertyMap.set should throw on receiving a list for a single valued property

### LINKED ISSUE #512: [css-typed-om] StylePropertMap.set for non-list properties
What should StylePropertyMap.set do for non-list properties when given multiple arguments? e.g.

```
styleMap.set('width', CSS.px(1), CSS.px(2), CSS.px(3));
styleMap.getAll('width'); // ???
```

According to the [spec](https://drafts.css-houdini.org/css-typed-om-1/#set-a-value-on-a-stylepropertymap), the "property model" will contain an entry `("width", [1px, 2px, 3px])`. It's then unclear what `getAll` should return. If we follow the [spec](https://drafts.css-houdini.org/css-typed-om-1/#get-all-values-from-a-stylepropertymap) strictly, we return `[[1px, 2px, 3px]]` (a single-element list containing a 3-element list), which is probably not right... Should `set` throw an exception or just set the first element?

nainar:
Fixes #512 

@tabatkins PTAL?

### DIFF
--- a/css-typed-om/Overview.bs
+++ b/css-typed-om/Overview.bs
@@ -354,9 +354,12 @@ The <dfn method for=StylePropertyMap>append(DOMString <var>property</var>,
 
     5. Let |values to set| be an empty list.
 
-    6. For each |value| in |values| if the [=algorithm that coerces value into an appropriate type for a given property=] does not throw an error, append the returned object to |values to set|.
+    6. If |property| is a single-valued property and |values| has more than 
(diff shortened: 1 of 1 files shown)


tabatkins:
Rebase this on latest - it currently reverts another commit accidentally.

nainar:
You have fixed this by adding the line here: https://github.com/w3c/css-houdini-drafts/blob/master/css-typed-om/Overview.bs#L380

Closing this PR

Issue #482, opened 2017-10-16
[css-typed-om] What should CSSNumericValue.type() do?


darrnshn:
Currently the IDL just says `??? type()` and there's an issue saying " Figure out how we want to represent the type of an expression in JS, and define the type() method accordingly.", but there's no GitHub issue associated with it.

@shans Seems like we already have wording for the type of a numeric value, but just need to decide how to represent it in Javascript?

tabatkins:
Yeah, the abstract "type" of a numeric value is a map of units to powers. We could simply reflect that, but it's not clear how useful it is.

nainar:
Seems like I suck as a scribe. The resolution should have read 

RESOLVED: type() will be a map of types to powers. @tabatkins as FYI.

darrnshn:
Thanks @nainar. @tabatkins are we also going to expose the [percent hint](https://drafts.css-houdini.org/css-typed-om-1/#cssnumericvalue-percent-hint)? Not that there's much use for it.

tabatkins:
Yeah, pretty much all the algos working on types need to look at the percent-hint; if anything in userland wants to work with types, it'll need it too. We shoudl just expose it as another key in the map, like "percent-hint".

darrnshn:
Fixed in #546

PR #546, opened 2017-12-17
Add wording for CSSNumericValue.type()

darrnshn:
issue #482 
Hi @tabatkins, PTAL. I'm not sure if I'm doing this right, so I only wrote up the interface for now. I'll do the actual text once we've agreed on the interface.

### DIFF
--- a/css-typed-om/Overview.bs
+++ b/css-typed-om/Overview.bs
@@ -647,13 +647,13 @@ are represented by subclasses of the {{CSSNumericValue}} interface.
     };
 
     dictionary CSSNumericType {
-        required long length;
-        required long angle;
-        required long time;
-        required long frequency;
-        required long resolution;
-        required long flex;
-        required long percent;
+        long length;
+        long angle;
+        long time;
+        long frequency;
+        long resolution;
+        long flex;
+        long percent;
         CSSNumericBaseType percentHint;
     };
 
--- a/css-typed-om/Overview.bs
+++ b/css-typed-om/Overview.bs
@@ -675,9 +675,6 @@ are represented by subclasses of the {{CSSNumericValue}} interface.
     };
 </xmp>
 
-Issue(482): Figure out how we want to represent the type of an expression in JS,
-and define the type() method accordingly.
-
 The methods on the {{CSSNumericValue}} superclass
 represent operations that all numeric values can perform.
 
@@ -1009,6 +1006,24 @@ The following are the arithmetic operations you can perform on dimensions:
     7. Return a new {{CSSMathSum}} object
         whose {{CSSMathSum/values}} internal slot
         is set to |result|.
+</div>
+
+<div algorithm="CSSNumericValue.type()">
+    The <dfn method for=CSSNumericValue>type()</dfn> method
+    returns a representation of the [=type=] of |this|.
+
+    When called, it must perform the following steps:
+
+    1. Let |result| be a new {{CSSNumericType}} with all attributes set to `null`.
+
+    2. For each |baseType| → |power| in the [=type=] of |this|,
+        1. If |power| is not 0,
+            set |result|'s |baseType| attribute to |power|.
+
+    3. If the [=percent hint=] of |this| is not null,
+        1. Set {{CSSNumericType/percentHint}} to the [=percent hint=] of |this|.
+
+    4. Return |result|.
 </div>
 
 <div algorithm="sum value">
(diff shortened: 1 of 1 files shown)


darrnshn:
@tabatkins Can you PTAL at the proposed IDL when you have time?

darrnshn:
Note that whether the base types should be required or optional depends on whether we need to differentiate between the types `{ length: 0 }` and `{}`. I think in the [original IRC log](https://github.com/w3c/css-houdini-drafts/issues/482#issuecomment-343238938), you wanted to differentiate between e.g. `1px/1px` and `1`? How important is this? IMO it's much easier to implement arithmetic if we don't have to differentiate between a "null" exponent and a zero exponent.

I don't think this has any observable impact on arithmetic, so only `CSSNumericValue.type()` can tell the difference between a null and a zero exponent.

Maybe @shans can weigh in on this as well since I asked him a similar question a while back.

tabatkins:
Nah, I can't find why I would have wanted to distinguish a zero exponent from a null exponent.  All the algorithms are careful to only talk about non-zero exponents.

And I'd probably want `.type()` to only emit the non-zero exponents, too.

darrnshn:
Thanks, I think @shans mentioned that animations distinguish between zero/null or something, but I don't know how useful that is for Typed OM.

Anyway, it looks like exponents should be optional either way. We can always distinguish zero/null later with no impact on the API.

tabatkins:
Yeah, we want to preserve zero *values* in a calc(), so the "shape" remains stable when animating across 0, but that reasoning doesn't apply to zero exponents in the type; those don't change in animations. ^_^

css-typed-om/Overview.bs
```diff
@@ -990,6 +1008,24 @@ The following are the arithmetic operations you can perform on dimensions:
         is set to |result|.
 </div>
 
+<div algorithm="CSSNumericValue.type()">
+    The <dfn method for=CSSNumericValue>type()</dfn> method
+    returns a representation of the [=type=] of |this|.
```
darrnshn: Might be better ways of saying this.

css-typed-om/Overview.bs
```diff
@@ -990,6 +1008,24 @@ The following are the arithmetic operations you can perform on dimensions:
...
+
+    When called, it must perform the following steps:
+
+    1. Let |result| be a new {{CSSNumericType}} with all attributes set to `null`.
```
darrnshn: Can I omit the "with all attributes set to null" since it's probably the default behaviour?

css-typed-om/Overview.bs
```diff
@@ -990,6 +1008,24 @@ The following are the arithmetic operations you can perform on dimensions:
...
+
+    2. For each |baseType| → |power| in the [=type=] of |this|,
+        1. If |power| is not 0,
+            set |result|'s |baseType| attribute to |power|.
```
darrnshn: Not sure if `attribute` is the correct term. Maybe `slot`?

css-typed-om/Overview.bs
```diff
@@ -990,6 +1008,24 @@ The following are the arithmetic operations you can perform on dimensions:
...
+
+    3. If the [=percent hint=] of |this| is not null,
+        1. Set {{CSSNumericType/percentHint}} to the [=percent hint=] of |this|.
```
darrnshn: This is a bit handwavy as it doesn't quite type check - `CSSNumericType.percentHint` is an enum whereas the `percent hint` of `this` is a string (I think).

css-typed-om/Overview.bs
tabatkins: nah, seems fine

css-typed-om/Overview.bs
tabatkins: Yeah.

css-typed-om/Overview.bs
tabatkins: The term for dictionaries is "member".  Alternately, since per WebIDL you're allowed to treat dictionaries as ordered maps, you can just say "set |result|[|baseType|] to |power|".

css-typed-om/Overview.bs
tabatkins: I think that's fine - enums turn into strings when they go to ECMAScript anyway.

darrnshn:
Cool, made the requested changes. PTAL

Issue #522, opened 2017-11-28
[css-typed-om] Normalizing 'top' as CSSPositionValue


darrnshn:
Following the [algorithm](https://drafts.css-houdini.org/css-typed-om-1/#positionvalue-normalization), `top` is a single keyword, so we go into [this case](https://drafts.css-houdini.org/css-typed-om-1/#from-single-keyword) with value `top` and bias `horizontal`. With these parameters, we would skip all the if statements and land on the else (step 6): If value is not the keyword "center", then this process fails. Since `top` is not `center`, we fail.

Did I interpret this right? I'm guessing this is not intended...

tabatkins:
This is definitely wrong, but I'm not sure exactly what it's trying to do; this algo is written too algorithmically for me.

It definitely *should* be treated the same as `top center`, tho.  Just need to see if any other cases need similar fixups or what.

darrnshn:
After staring at this for a bit, the algorithm might work if we initialize the bias to `auto` when handling the single value case.

PR #557, opened 2018-01-12
Change CSSPositionValue normalization to use auto bias

darrnshn:
Should fix issue #522 
@tabatkins PTAL when you have time


tabatkins:
I don't wanna think about it because I don't like the algorithm, but if it returns the right results in your tests when you code it that way, I'm happy to accept it. ^_^

PR #559, opened 2018-01-14
[css-typed-om] Add shortcuts for other arithmetic operations

### LINKED ISSUE #503: [css-typed-om] Simplifying expressions in CSSMathValue subclasses.
In [the algorithm for adding CSSNumericValues](https://drafts.css-houdini.org/css-typed-om-1/#dom-cssnumericvalue-add), we have an explicit step (Step 3) for simplifying the expression if all values are `CSSUnitValues` with the same unit. However, none of the other subclasses seem to have this step, even though theoretically it should be possible (e.g. `min(2px, 3px) == 2px`). The inconsistency is a bit confusing and makes addition a special case in terms of implementation. Is there any particular reason for making addition simplify but not the others?

darrnshn:
Resolves #503 
@tabatkins PTAL. Wording for `mul` might be a bit awkward.

### DIFF
--- a/css-typed-om/Overview.bs
+++ b/css-typed-om/Overview.bs
@@ -725,7 +725,7 @@ The following are the arithmetic operations you can perform on dimensions:
         with the same {{CSSUnitValue/unit}} internal slot as |this|,
         and a {{CSSUnitValue/value}} internal slot set to the negation of |this|’s.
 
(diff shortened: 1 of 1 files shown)


tabatkins:
Eh, it's fine. Thanks for the edit!

PR #563, opened 2018-01-15
[css-typed-om] Restrict unsupported CSSStyleValues to the property that it was parsed with

### LINKED ISSUE #553: [css-typed-om] Behaviour of unsupported CSSStyleValues
When you get an unsupported `CSSStyleValue`, does the spec have any requirements about its behaviour? I couldn't find any wording on this. I would think that at the minimum, we should be able to use them correctly in `StylePropertyMap`s.

```
const value = styleMap.get('filter');
otherStyleMap.set('filter', value);
```

darrnshn:
Resolves #553 
@tabatkins PTAL, thanks!

### DIFF
--- a/css-typed-om/Overview.bs
+++ b/css-typed-om/Overview.bs
@@ -124,8 +124,10 @@ interface CSSStyleValue {
 </pre>
 
 {{CSSStyleValue}} objects are the base class of all CSS Values accessible via the Typed OM API.
+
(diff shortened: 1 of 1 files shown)


tabatkins:
This'll need more clarity eventually, but we need to tighten up that whole section anyway. This works for now. :+1:

Issue #584, opened 2018-01-28
[css-typed-om] "algorithm that coerces value into an appropriate type for a given property" calls "parse a CSSStyleValue" with missing arg


gsnedders:
>      If value is a DOMString, Parse a CSSStyleValue with property property and value value and return the resulting value. If the result is null, throw a TypeError and exit this algorithm. Otherwise, append each list-valued property iteration in the result to a values to append object and return values to append.

This doesn't state whether `parseMultiple` is true/false.

dbaron:
Same as #567?

tabatkins:
Yup, and fixed.

PR #585, opened 2018-01-28
Invoke parsing algorithm correctly in style value coercion text

### LINKED ISSUE #567: [css-typed-om] algorithm for coercing value should specify parseMultiple flag
The spec describes an [algorithm that coerces value into an appropriate type for a given property](https://drafts.css-houdini.org/css-typed-om-1/#algorithm-that-coerces-value-into-an-appropriate-type-for-a-given-property).

It says:

> Parse a CSSStyleValue with property property and value value and return the resulting value.

I think it should also say that the parseMultiple flag should be true.  And it should describe the value as a cssText, so I think it should instead say something more like:

> Parse a CSSStyleValue with property property, cssText value, and parseMultiple true, and return the resulting value.

darrnshn:
Closes #567 and #584 

Hi @tabatkins, PTAL


tabatkins:
Fixed while I was Infra-ifying these algos.

PR #591, opened 2018-01-29
Clarify CSSNumericValues.equals

### LINKED ISSUE #576: [css-typed-om] definition of CSSMathValue.equals() should be a little clearer
The [definition of CSSMathValue.equals()](https://drafts.css-houdini.org/css-typed-om-1/#dom-cssnumericvalue-equals) seems to lack a bit of formality in the final step.  In particular, the final step currently begins with:

> If all items in thisAndValues are of type CSSMathValue:

and then has two options that require iterating over a bunch of values.

The text doesn't actually say how to take this iteration and produce a single result (true or false) from the equals function.  It probably should.

It also feels like the wording could be clearer in other ways, e.g., in terms of describing what exactly is being iterated over and how many recursive calls to the algorithm are made.

darrnshn:
Fixes #576 . Tries to clarify `CSSNumericValue.equals` a bit by pulling pairwise equality into its own algo.  Not quite sure if this is idiomatic bikeshed.

Hi @tabatkins PTAL

### DIFF
--- a/css-typed-om/Overview.bs
+++ b/css-typed-om/Overview.bs
@@ -863,20 +863,35 @@ The following are the arithmetic operations you can perform on dimensions:
 
     1. Replace each [=list/item=] of |values|
         with the result of [=rectifying a numberish value=] for the [=list/item=].
-
-    2. Let |thisAndValues| be |this| prepended to |values|.
-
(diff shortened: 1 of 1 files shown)


tabatkins:
Heh, I started writing this myself, then realized our solutions were almost identical. I'll merge yours and fiddle a bit. ^_^

PR #592, opened 2018-01-29
StylePropertyMap's model is sequence<CSSStyleValue> pairs

### LINKED ISSUE #571: [css-typed-om] "get all values from a StylePropertyMap" assumes underlying model might not be list
Typed OM defines an [algorithm to get all values from a `StylePropertyMap`](https://drafts.css-houdini.org/css-typed-om-1/#get-all-values-from-a-stylepropertymap).  This says:

> 3. If StylePropertyMap’s property model contains an entry for property:
>
>     1. If property is a list-valued property, return the list of values assigned to that property.
>
>     2. If property is not a list-valued property, let values initially be the empty list and append the value assigned to the property to this list and return values.

However, earlier the spec [says](https://drafts.css-houdini.org/css-typed-om-1/#the-stylepropertymap):

> A StylePropertyMapReadOnly object has an associated property model, which is a list of property - sequence&lt;CSSStyleValue> pairs.

Given this, it's not clear why this distinction between list-valued and non-list-valued properties needs to be made; they're all represented as `sequence<CSSStyleValue>`.

nainar:
Fixes #571

### DIFF
--- a/css-typed-om/Overview.bs
+++ b/css-typed-om/Overview.bs
@@ -337,15 +337,8 @@ probably in an appendix.
     2. If |property| is not a [=supported property name=],
         [=throw=] a {{TypeError}} and exit this algorithm.
   
-    3. If {{StylePropertyMap}}’s [=contained properties map=] contains an entry for |property|:
-        
-        1. If |property| is a [=list-valued property=],
-            return the list of |values| assigned to that property.
-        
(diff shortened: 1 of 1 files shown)


tabatkins:
Ah, took care of this as a result of Infra-ifying things.

Issue #520, opened 2017-11-20
[css-typed-om] Should the product of two unit maps be addition instead of union?


darrnshn:
In the spec (Step 2.3.1.1 for CSSMathProduct in the [sum value section](https://drafts.css-houdini.org/css-typed-om-1/#sum-value)) for multiplying two unit maps, the wording says to take the union of the unit maps. I might be misunderstanding unit maps, but wouldn't that make something like `1px * 1px == { px -> 1 }` instead of `1px * 1px == { px -> 2 }`? Should we add the entries/exponents in the two unit maps instead (like in types)?

tabatkins:
Ah, that's what's intended by "union" (result should be `{px -> 2}`), but I agree that it's unclear as written.

I should be explicitly linking to [multiplying types](https://drafts.css-houdini.org/css-typed-om-1/#cssnumericvalue-multiply-two-types).

darrnshn:
Fixed by #593

PR #593, opened 2018-01-29
Clarify the union of two unit maps

### LINKED ISSUE #578: [css-typed-om] need to define (or link to?) concept of unioning unit maps
The [rules for creating a sum value](https://drafts.css-houdini.org/css-typed-om-1/#create-a-sum-value) from a `CSSNumericValue` have wording in the part about processing a `CSSMathProduct` that says:

> Let item be a tuple with its value set to the product of the values of item1 and item2, and its unit map set to the union of the unit maps of item1 and item2, with all entries with a zero value removed.

The concept of how to union unit maps isn't defined (or at least it isn't a link here), and I didn't see a definition where unit maps are defined.

It seems like unioning unit maps should merge identical units, for example unioning `["px" → 1]` and `["px" → 2]` should yield `["px" → 3]`.  Likewise, unioning `["px" → 1]` and `["px" → -1]` should yield `[]`.   I don't think those follow from a straightforward interpretation of the word "union", but I think they are what is expected.  (I think the straightforward interpretation of "union" does work when the types are different, though.)

Given that, I also wonder if the concept should be called multiplying rather than unioning.

darrnshn:
Fixes #578 and #520 by defining the product of two unit maps separately.

Hi @tabatkins , PTAL

### DIFF
--- a/css-typed-om/Overview.bs
+++ b/css-typed-om/Overview.bs
@@ -1105,7 +1105,7 @@ The following are the arithmetic operations you can perform on dimensions:
                             1. Let |item| be a [=tuple=] with its [=sum value/value=]
                                 set to the product of the [=sum value/values=] of |item1| and |item2|,
                                 and its [=sum value/unit map=]
-                                set to the union of the [=sum value/unit maps=] of |item1| and |item2|,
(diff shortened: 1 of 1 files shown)


tabatkins:
Merged, but we should probably explicitly remove zero values from the result (both of this and the "multiple types" algo).

PR #613, opened 2018-01-30
Use defaults for CSSMatrixComponentOptions

### LINKED ISSUE #612: [css-typed-om] specify CSSMatrixComponentOptions default with IDL defaults?
The [`CSSMatrixComponent` constructor spec](https://drafts.css-houdini.org/css-typed-om-1/#dom-cssmatrixcomponent-cssmatrixcomponent) says:

> If options was passed and has a CSSMatrixComponentOptions field, set this’s is2D internal slot to the value of that field.

One problem with this is that CSSMatrixComponentOptions probably meant to say is2D.

However, it seems like the normal way to specify these things is using WebIDL dictionary defaults.  Maybe that's better, and it could avoid needing to write the part of this sentence with the mistake?

darrnshn:
Fixes #612 . Changes behaviour: you can no longer pass in an option with null `is2D`.

Hi @tabatkins , PTAL

### DIFF
--- a/css-typed-om/Overview.bs
+++ b/css-typed-om/Overview.bs
@@ -2061,7 +2061,7 @@ is a [=list=] of {{CSSTransformComponent}}s.
     };
 
     dictionary CSSMatrixComponentOptions {
-        required boolean is2D = false;
+        boolean is2D = false;
     };
 </xmp>
 
--- a/css-typed-om/Overview.bs
+++ b/css-typed-om/Overview.bs
@@ -2061,7 +2061,7 @@ is a [=list=] of {{CSSTransformComponent}}s.
     };
 
     dictionary CSSMatrixComponentOptions {
-        boolean is2D;
+        required boolean is2D = false;
     };
 </xmp>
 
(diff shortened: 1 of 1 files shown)


tabatkins:
Why'd you make it required? That kinda clashes with providing a default value for it. ^_^

darrnshn:
Oh oops, I thought `required` would make `is2D` not nullable. Fixed.

tabatkins:
It was never nullable (that would have required a `boolean?`), it just could have been not passed before; adding a default value prevents that.

darrnshn:
@tabatkins Hi Tab, it appears that this changes behaviour in a way that I didn't expect.

Before, we had two possibilities:
1. Don't pass in `options`. `is2D` is inferred from `matrix.is2D`.
2. Pass in `options`. `is2D` is set to `options.is2D`.

It seems that with this change, the first possibility changes to:
1. Don't pass in `options`. `is2D` is `false` (the default value).

I didn't expect this because I thought `options` was an `optional` argument, we would still have the "infer" code path, but it looks like IDL magic has made it impossible.

Is this the intended behaviour? Or is this a bug with the Blink IDL?

tabatkins:
Ah, you're right, that is a behavior change. We should revert this and just address the issue more directly, then.

PR #615, opened 2018-01-31
Clarify when CSSTransformValue.toMatrix() throws TypeError

### LINKED ISSUE #608: [css-typed-om] have a note saying when CSSTransformValue.toMatrix() doesn't work
The definition of [CSSTransformValue.toMatrix()](https://drafts.css-houdini.org/css-typed-om-1/#dom-csstransformvalue-tomatrix) has technical prose:

> As the entries of such a matrix are defined relative to the px unit, if any &lt;length>s in func involved in generating the matrix are not compatible units with px, throw a TypeError.

that says that the function will sometimes throw.  But it's probably worth adding a note to make this clearer: it will happen when there is use of [relative lengths](https://drafts.csswg.org/css-values-4/#relative-lengths) or [percentages](https://drafts.csswg.org/css-transforms/#transform-functions) in transform functions.

nainar:
Fixes #608

### DIFF
--- a/css-typed-om/Overview.bs
+++ b/css-typed-om/Overview.bs
@@ -80,6 +80,7 @@ spec:css-syntax-3;
 spec:css-values-4;
     type:type; text:<position>
     type:value; text:in
+    type:dfn; text:relative length
 spec:infra;
     type: dfn;
         text: list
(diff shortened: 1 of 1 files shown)


tabatkins:
Manually folded into https://github.com/w3c/css-houdini-drafts/commit/72d9e39365331e13f002bb3e3a4a5fa3753787df

PR #621, opened 2018-02-01
Extend match concept to other style values

### LINKED ISSUE #566: [css-typed-om] algorithm for coercing value doesn't describe how to check grammar
The spec describes an [algorithm that coerces value into an appropriate type for a given property](https://drafts.css-houdini.org/css-typed-om-1/#algorithm-that-coerces-value-into-an-appropriate-type-for-a-given-property).

It says:

> If value is a CSSStyleValue,
>
> If value does not match the grammar of a list-valued property iteration of property, throw a TypeError and exit this algorithm. Otherwise, return the value.

But it doesn't say how to tell if a `CSSStyleValue` matches the grammar for a list-valued property iteration of the property.

darrnshn:
Fixes #566 ? Not quite sure how to do this. How a "CSS production" matches a "grammar" is a bit handwavy.

Hi @tabatkins , PTAL.

### DIFF
--- a/css-typed-om/Overview.bs
+++ b/css-typed-om/Overview.bs
@@ -355,7 +355,7 @@ probably in an appendix.
     and a [=string=] or {{CSSStyleValue}} |value|:
 
     : If |value| is a {{CSSStyleValue}},
-    :: If |value| does not match the grammar of a [=list-valued property iteration=] of |property|,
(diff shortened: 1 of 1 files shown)


tabatkins:
A reasonable start, I'll merge and build on this.

PR #639, opened 2018-02-05
Clarify CSSURLImageValue.url

### LINKED ISSUE #630: [css-typed-om] CSSURLImageValue constructor probably shouldn't require "url()" around the URL
As specified today, the [`CSSURLImageValue` constructor] seems to require that `url()` be around the URL in the string passed to it, since it depends on the `<url>` production.

I suspect it shouldn't require `url()`... but if that was the intent, an example should be given.

The spec should also specify more clearly how the parameter is processed, for example, whether CSS character escapes are processed.

darrnshn:
Fixes #630 .

Hi @tabatkins PTAL, not sure what the correct wording should be.

### DIFF
--- a/css-typed-om/Overview.bs
+++ b/css-typed-om/Overview.bs
@@ -2650,11 +2650,11 @@ objects, the {{CSSURLImageValue/url}} attribute contains the URL that references
     when called,
     perform the following steps:
 
-    1. If the |url| passed into the constructor doesn't correctly
-        parse as a <<url>>, throw a {{TypeError}} and exit this algorithm.
+    1. If the |url| passed into the constructor is not a valid url,
+        throw a {{TypeError}} and exit this algorithm.
     2. Else, return a new {{CSSURLImageValue}}
         with its {{CSSURLImageValue/url}} internal slot
-        set to |url|.
+        set to |url| with CSS character escapes processed.
 </div>
 
 Issue(159): Spec up ColorValue
--- a/css-typed-om/Overview.bs
+++ b/css-typed-om/Overview.bs
@@ -2743,12 +2743,11 @@ objects, the {{CSSURLImageValue/url}} attribute contains the URL that references
     when called,
     perform the following steps:
 
-    1. If the |url| passed into the constructor is not a valid url,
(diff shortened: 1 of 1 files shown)


css-typed-om/Overview.bs
```diff
@@ -2650,11 +2650,11 @@ objects, the {{CSSURLImageValue/url}} attribute contains the URL that references
     when called,
     perform the following steps:
 
-    1. If the |url| passed into the constructor doesn't correctly
-        parse as a <<url>>, throw a {{TypeError}} and exit this algorithm.
+    1. If the |url| passed into the constructor is not a valid url,
```
darrnshn: Can I link to some definition of valid url?

css-typed-om/Overview.bs
```diff
@@ -2650,11 +2650,11 @@ objects, the {{CSSURLImageValue/url}} attribute contains the URL that references
...
+        throw a {{TypeError}} and exit this algorithm.
     2. Else, return a new {{CSSURLImageValue}}
         with its {{CSSURLImageValue/url}} internal slot
-        set to |url|.
+        set to |url| with CSS character escapes processed.
```
darrnshn: Uh not sure what to do here...

css-typed-om/Overview.bs
```diff
@@ -2743,7 +2743,9 @@ objects, the {{CSSURLImageValue/url}} attribute contains the URL that references
     when called,
     perform the following steps:
 
-    1. Return a new {{CSSURLImageValue}}
+    1. If the |url| passed into the constructor is not a [[url#valid-url-string]],
```
darrnshn: I couldn't figure out how to link the URL spec correctly sorry :/

annevk: That doesn't really work. You need to use the URL parser (`<a>URL parser</a>`) and make sure the return value is not failure. However, you'll also need to pass a base URL for that. And it seems you want to treat fragments as a special value per issue discussion so this probably needs some more thought.

tabatkins:
Closing due to substantial redesign, see <https://github.com/w3c/css-houdini-drafts/issues/716#issuecomment-368656793>

PR #647, opened 2018-02-12
Serialize out of range values as calcs

### LINKED ISSUE #574: [css-typed-om] What does lack of range restriction imply about serialization?
The [section on CSSNumericValue](https://drafts.css-houdini.org/css-typed-om-1/#numeric-objects) says:

> CSSNumericValue objects are not range-restricted. Any valid numeric value can be represented by a CSSNumericValue, and that value will not be clamped, rounded, or rejected when set on a declared StylePropertyMap or inline StylePropertyMap. Instead, clamping and/or rounding will occur during computation of style.

(It then proceeds to give an example demonstrating this.)

What this doesn't seem to address is what implications this has for CSS's existing object model APIs that involve serialization.  It's long been a principle of these APIs that they always give output that is valid CSS.  This means that, for example, getting `element.style.cssText`, modifying it, and setting it back to `element.style.cssText` should change only the thing that was modified.  Likewise for `getPropertyValue` and `setProperty`, or the `element.style.propertyName` that is equivalent to them.  However, the rules for these setters require following the rules of CSS parsing and dropping any declarations that are not valid.  This means that serializing things that don't round-trip correctly would break anything that depends on this round-tripping.

It seem like there are a bunch of bad options here (listed in the order that seems best to worst to me at first glance, although I'm sure I'm missing many considerations):
* serialize CSSUnitValue that are out-of-range as `calc()` expressions
* not do the thing quoted above, but instead do rejection of `CSSNumericValue`s when they're set in some cases
* serialize all CSSUnitValue as `calc()` expressions
* break round-tripping as the spec currently does

I tried to look through the spec to see if one of these options was specified later, but I didn't see anything.  If there is something I missed, it would be good to have a pointer to it from this section.

It's also entirely possible that the WG has discussed this and I've forgotten (or missed it).

nainar:
Fixes #574

### DIFF
--- a/css-typed-om/Overview.bs
+++ b/css-typed-om/Overview.bs
@@ -360,7 +360,9 @@ probably in an appendix.
     : If |value| is a {{CSSStyleValue}},
     :: If |value| does not [=match a CSS production|match the grammar=] of a [=list-valued property iteration=] of |property|,
         [=throw=] a {{TypeError}} and exit this algorithm.
-        Otherwise, return the |value|.
+        If the |property| only allows certain values, and is passed a {{CSSUnitValue}} |value| that is outside that range, 
+        wrap |value| in a {{CSSMathSum}}, with only one entry in |values|, so that it serializes to 'calc(' + |value| + ')'.
+        Return the |value|.
 
     : If |value| is a {{DOMString}},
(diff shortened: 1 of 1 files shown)


plinss:
When this gets merged (or closed), please ping me so I can clean up the mercurial side of the branch

tabatkins:
@plinss You can clean up the mercurial side now, if you haven't already.

PR #673, opened 2018-02-15
Remove StylePropertyMap.update()

### LINKED ISSUE #656: [css-typed-om-1]update() behaves a bit oddly for non-single-valued properties
https://drafts.css-houdini.org/css-typed-om-1/#dom-stylepropertymap-update-property-updatefn is defined to do a get(), then pass the single CSSStyleValue to the given function, then do a set().

For a non-single-valued property this would lose all but the first item in the list if an identity UpdateFunction is used.  This seems like slightly odd behavior.

nainar:
Fixes #656

### DIFF
--- a/css-typed-om/Overview.bs
+++ b/css-typed-om/Overview.bs
@@ -241,13 +241,10 @@ interface StylePropertyMapReadOnly {
     stringifier;
 };
 
-callback UpdateFunction = CSSStyleValue (CSSStyleValue oldValue);
-
 [Exposed=Window]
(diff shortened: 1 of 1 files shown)


plinss:
@nainar In the future, please don't create PRs in the main repo. Fork the repo to your own account and generate the PRs from there. Having all the extra branches in the main repo puts an additional burden on the CI server at drafts.css-houdini.org as it runs bikeshed on each active branch.

nainar:
Sorry, will do from now on.

plinss:
Thanks!

Issue #667, opened 2018-02-15
[css-typed-om-1] CSSVariableReferenceValue's "variable" getter needs to be defined.


bzbarsky:
Presumably it returns the "variable" internal slot's value?

darrnshn:
I'm guessing you meant CSSVariableReferenceValue?

bzbarsky:
Er, yes, I do.  They're kinda lumped together in a way that confused me.  ;(

tabatkins:
Fixed in #676.

(I left it out on the assumption that, if undefined, it just does the normal thing.  Specs generally leave off *both* getters and setters for attributes that just reflect the internal slot.)

Issue #666, opened 2018-02-15
[css-typed-om-1] CSSUnparsedValue's length could be defined more clearly


bzbarsky:
It could be defined to just return the length of the list in [[tokens]].

tabatkins:
Fixed in #676

PR #676, opened 2018-02-16
Clarify CSSUnparsedValue and CSSVariableReferenceValue.

### LINKED ISSUE #663: [css-typed-om-1] Behavior of indexed setter on CSSUnparsedValue is not defined.
Per https://heycam.github.io/webidl/#idl-indexed-properties 

>  If the operation used to declare the indexed property setter did not have an identifier, then the interface definition must be accompanied by a description of how to set the value of an existing indexed property and how to set the value of a new indexed property for a given property index and value.

but there's nothing like this in the Typed OM draft.

The getter behavior is not really defined either (because there is no decription for how to "determine the value of an indexed property"), but it can at least be guessed at.

darrnshn:
Should fix #663 #666 #667 

Hi @bzbarsky , PTAL? The language for indexed getter/setter is still a bit handwavy :P Thanks!


tabatkins:
Works for now. :+1:

Issue #678, opened 2018-02-16
 [css-typed-om-1]Probably need to define the order of multiplications in mul() step 3 and step 4


bzbarsky:
https://drafts.css-houdini.org/css-typed-om-1/#dom-cssnumericvalue-mul step 3 multiplies together a bunch of floating-point numbers.  Multiplication on floating-point numbers is not associative, so the answer depends on how you do it, exactly.  Ideally this would be specified.

bzbarsky:
Same thing for step 4 of mul.

tabatkins:
Fixed.

PR #685, opened 2018-02-16
Enforce ordering on order of operations

### LINKED ISSUE #677: [css-typed-om-1]Probably need to define the order of additions in add() step 3
https://drafts.css-houdini.org/css-typed-om-1/#dom-cssnumericvalue-add step 3 says:

> If all of the items in values are CSSUnitValues and have the same unit, return a new CSSUnitValue whose unit internal slot is set to this’s unit internal slot, and value internal slot is set to the sum of the value internal slots of the items in values.

Unfortunately you are adding doubles here.  If there are more than 3 things in the list, you have to deal with the fact that floating-point addition is not associative and that the answer will depend on the order you perform your additions in.  Ideally this spec would define the order.

darrnshn:
Fixes #677 and #678 .
Hi @tabatkins PTAL

### DIFF
--- a/css-typed-om/Overview.bs
+++ b/css-typed-om/Overview.bs
@@ -899,7 +899,7 @@ The following are the arithmetic operations you can perform on dimensions:
         whose {{CSSUnitValue/unit}} internal slot is set to |this|’s {{CSSUnitValue/unit}} internal slot,
         and {{CSSUnitValue/value}} internal slot is set to the sum of
         the {{CSSUnitValue/value}} internal slots
-        of the [=list/items=] in |values|.
+        of the [=list/items=] in |values| (from left to right).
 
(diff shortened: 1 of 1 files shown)


css-typed-om/Overview.bs
```diff
@@ -899,7 +899,7 @@ The following are the arithmetic operations you can perform on dimensions:
         whose {{CSSUnitValue/unit}} internal slot is set to |this|’s {{CSSUnitValue/unit}} internal slot,
         and {{CSSUnitValue/value}} internal slot is set to the sum of
         the {{CSSUnitValue/value}} internal slots
-        of the [=list/items=] in |values|.
+        of the [=list/items=] in |values| (from left to right).
```
bzbarsky: Does "from left to right" mean "a + (b + c)" or "(a + b) + c"?  It would be better to just make this explicit instead of trying to be terse.

PR #688, opened 2018-02-19
computedStyleMap() must return same object

### LINKED ISSUE #660: [css-typed-om-1] Does computedStyleMap() always return a new object?
https://drafts.css-houdini.org/css-typed-om-1/#computed-stylepropertymapreadonly-objects

If so, it should be annotated [NewObject].

That said, why would it?  There's no reason for it too, afaict, just like attributeStyleMap and styleMap don't.

darrnshn:
Fixes #660 . 

Hi @tabatkins PTAL. Not sure how to write it explicitly in the algorithm (how to "cache" it?)


tabatkins:
To cache, we specify that Element has a new private slot, and on first call of the method, it stores the object into that slot before returning it; later calls just return the value from the slot.

darrnshn:
Thanks, I've added explicit caching logic, PTAL?

PR #706, opened 2018-02-22
Add a few examples to the reification table

darrnshn:
Still drafting this up to get a feel for this, @tabatkins PTAL?

### DIFF
--- a/css-typed-om/Overview.bs
+++ b/css-typed-om/Overview.bs
@@ -3574,6 +3574,11 @@ which will be filled in with more detail as we go.
 
 : <a>align-content</a>
 ::
+    1. If the specified or computed value is the ''normal'' keyword,
+        return the result of reifying the "normal" keyword.
+    2. If the value is a keyword from <<content-distribution>>,
+        return the result of reifying the keyword value.
+    3. Otherwise, the value is not supported.
 
 : <a>align-items</a>
 ::
@@ -4801,6 +4806,11 @@ which will be filled in with more detail as we go.
 
 : <a>top</a>
 ::
+    1. If the specified or computed value is the ''auto'' keyword,
+        return the result of reifying the "auto" keyword.
+    2. If the specified or computed value is of type <<length>>,
+        return the result of reifying the <<length>> value.
+    3. Otherwise, return the result of reifying the <<percentage>> value.
 
 : <a>transform-style</a>
 ::


css-typed-om/Overview.bs
```diff
@@ -3574,6 +3574,11 @@ which will be filled in with more detail as we go.
 
 : <a>align-content</a>
 ::
+    1. If the specified or computed value is the ''normal'' keyword,
```
darrnshn: I can either do this, or have separate tables for specified vs computed (will have duplications)?

css-typed-om/Overview.bs
```diff
@@ -3574,6 +3574,11 @@ which will be filled in with more detail as we go.
...
+        return the result of reifying the "normal" keyword.
+    2. If the value is a keyword from <<content-distribution>>,
```
darrnshn: Is this legit, or do I need to specify every keyword?

css-typed-om/Overview.bs
```diff
@@ -3574,6 +3574,11 @@ which will be filled in with more detail as we go.
...
+        return the result of reifying the keyword value.
+    3. Otherwise, the value is not supported.
```
darrnshn: Interesting case with like `<<baseline-position>>`, which can be 'baseline', 'first baseline', 'last baseline'. We can either reify 'baseline' as keyword and the others as unsupported, or treat all three as unsupported in case we promote `<<baseline-position>>` to its own value?

css-typed-om/Overview.bs
tabatkins: Depends on the property. align-content has the same behavior in specified and computed, so we can combine them.

css-typed-om/Overview.bs
tabatkins: This is legit.

css-typed-om/Overview.bs
tabatkins: We should reify `baseline` and `first baseline` both to the single keyword `baseline`, and treat `last baseline` as unsupported.

Same for `<content-position>`: reify if there's no `<overflow-position>`, unsupported otherwise.

css-typed-om/Overview.bs
```diff
@@ -4801,6 +4806,11 @@ which will be filled in with more detail as we go.
 
 : <a>top</a>
 ::
+    1. If the specified or computed value is the ''auto'' keyword,
+        return the result of reifying the "auto" keyword.
+    2. If the specified or computed value is of type <<length>>,
+        return the result of reifying the <<length>> value.
```
tabatkins: Need to link to the algo defining this (reify a numeric value). Can combine this with the next step, since lengths and %s use the same algo.

PR #726, opened 2018-03-06
Add some reification

darrnshn:
Hi @tabatkins PTAL.


tabatkins:
Looks great! I'm gonna add an actual defined term for reifying to a CSSStyleValue soon, but feel free to keep going as this is, I'll do a find-and-replace later.

PR #731, opened 2018-03-09
Add font reification

darrnshn:
Hi @tabatkins , PTAL. I got some questions too.

### DIFF
--- a/css-typed-om/Overview.bs
+++ b/css-typed-om/Overview.bs
@@ -3509,58 +3509,140 @@ regardless of what property it is for.
 ::
 
 : <a>font</a>
-::
+:: For both specified and computed values,
+    reify as a {{CSSStyleValue}}
+    and return the result.
 
 : <a>font-family</a>
-::
+:: For both specified and computed values,
+    reify as a {{CSSStyleValue}}
+    and return the result.
 
 : <a>font-language-override</a>
-::
+:: For both specified and computed values:
+
+    1. If the value is <css>normal</css>,
+        [=reify an identifier=] from the value
+        and return the result.
+
+    2. Otherwise, reify as a {{CSSStyleValue}}
+        and return the result.
 
 : <a>font-max-size</a>
-::
+:: For both specified and computed values:
+
+    1. If the value is an <<absolute-size>>, <<relative-size>> or <css>infinity</css>, 
+        [=reify an identifier=] from the value
+        and return the result.
+
+    2. Otherwise, [=reify a numeric value=] from the value
+        and return the result.
 
 : <a>font-min-size</a>
-::
+:: Same as 'font-size'
 
 : <a>font-optical-sizing</a>
-::
+:: For both specified and computed values,
+    [=reify an identifier=] from the value
+    and return the result.
 
 : <a>font-palette</a>
-::
+:: For both specified and computed values:
+
+    1. If the value is <css>normal</css>, <css>light</css> or <css>dark</css>,
+        [=reify an identifier=] from the value
+        and return the result.
+
+    2. Otherwise, reify as a {{CSSStyleValue}}
+        and return the result.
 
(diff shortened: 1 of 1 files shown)


css-typed-om/Overview.bs
```diff
@@ -3509,58 +3509,140 @@ regardless of what property it is for.
...
+        and return the result.
+
+    2. Otherwise, reify as a {{CSSStyleValue}}
+        and return the result.
 
 : <a>font-synthesis</a>
-::
+:: For both specified and computed values:
+
+    1. If the value is <css>none</css>, <css>weight</css>, <css>style</css> or <css>small-caps</css>,
```
darrnshn: The syntax is `none | [ weight || style || small-caps ]`. Is this correct? I wonder what will happen when we support values like `weight style`. Might be some backwards compat issues there?

css-typed-om/Overview.bs
```diff
@@ -3509,58 +3509,140 @@ regardless of what property it is for.
...
+        and return the result.
 
 : <a>font-variant</a>
-::
+:: For both specified and computed values,
+    reify as a {{CSSStyleValue}}
+    and return the result.
 
 : <a>font-variant-alternates</a>
 ::
```
darrnshn: Same issue as font-synthesis.

css-typed-om/Overview.bs
```diff
@@ -3509,58 +3509,140 @@ regardless of what property it is for.
...
 
 : <a>font-variant-emoji</a>
-::
+:: For both specified and computed values,
+    [=reify an identifier=] from the value
+    and return the result.
 
 : <a>font-variation-settings</a>
-::
+:: For both specified and computed values:
```
darrnshn: The syntax is `none | <something>#`, so it's neither list-valued nor single-valued. I think reification-wise it's fine, but the wording for `styleMap.append` etc. might be hard (can append `<something>` if it's not `none`, can never append `none`). Alternatively, we can always represent the list values as a separate type (like `transform` which suffers the same problem), but it means we can never `append` to it directly.

css-typed-om/Overview.bs
darrnshn: Specifically, I mean if we reify `weight style` as a list, then I have to write code like:

```
// Want to know if 'small-caps' is synthesized.
const result = styleMap.get('font-synthesis');
if (result is keyword with value 'small-caps' or result is a list containing value 'small-caps')
```

instead of just
```
if (result is a list containing value 'small-caps')
```

darrnshn:
Hi @tabatkins , can you PTAL?

css-typed-om/Overview.bs
tabatkins: Ah, I figured we probably had an example of this somewhere; the 'animation' handling of "none" seemed sufficiently weird that I was pretty sure *somewhere* wouldn't have gone for it.

css-typed-om/Overview.bs
tabatkins: No, "weight style" should definitely be a CSSStyleValue; it's a single value with two keywords in it, not a list-valued property.

css-typed-om/Overview.bs
tabatkins: (It's "normal", not "none".)

Let's leave this off for the moment - we can just reify to a CSSStyleValue until we figure it out. I suspect this'll be something we want to run by the group and get opinions on.

tabatkins:
Other than font-variation-settings, looks good.

css-typed-om/Overview.bs
darrnshn: Ah yes that makes sense. We can just rely on the upgrade mechanism here.

darrnshn:
Fixed `font-variation-settings` and added `font-variant-alternates`. PTAL thanks!

PR #770, opened 2018-06-22
Fix typo "auther" -> "author"

numaanashraf:
Fix typo in overview


numaanashraf:

Issue #315, opened 2016-09-30
[css-properties-values-api] var() references to <length> type properties with font-relative lengths


tim-loh:
Consider a property --font-size, registered to be of `<length>` type. What should the computed values of font-size and --font-size be?

--font-size: 10em;
font-size: var(--font-size);

If --font-size is substituted as "10em", then font-size is `10*parent` and --font-size is `100*parent`. If it is substituted after computation then what value do we use for the em unit?

tabatkins:
Oh dang, this is a difficult question.  I didn't anticipate this adding a new cyclic dependency.  (Good thing `currentcolor` is converted at used-value time, or else it'd cause the same problem with `color`.)

We'll need to make an arbitrary decision here.  I think my preferred solution is to lean on normal variable dependency-cycle resolution - custom properties containing a font-relative length (this requires them to be typed, with a `<length>` member) add an edge to the dependency graph marking them as depending on `font-size`, so in your example, the `var()` reference would be treated as invalid.

jyc:
Is it valid to conceptualize this as there being a phony variable `font-size` and have cycle detection treat all dimension tokens with font size relative lengths in variables whose set of possible syntaxes contains `<length>`, `<length-percentage>`, or `<transform-list>` (anything where we could have a length) like it would treat `var(font-size)`?

Should the following declarations with `--messy` being an unregistered custom property and`--some-transform-list` being registered with syntax `<transform-list>` and initial value `scale(1)` end up with the `font-size` inheriting and `--some-transform-list` being `""`?

It seems like because the `5em` is actually the argument to `scale` which takes a `<number>` rather than `perspective` which takes a `<length>`, if `var(--messy)` were substituted then the declaration of `--some-transform-list` would be syntactically invalid and `font-size` be computed as `16px`. However, if we are trying to detect cycles before substituting, then we can't know what parameter `5em` ends up as the argument for, and can only assume the worst.

```
--messy: ) scale(
--some-transform-list: perspective(var(--messy) 5em);
font-size: var(--some-transform-list, 16px);
```

jyc:
Also, what about `rem`s? Should these only create a cycle when we're at the root?

jyc:
Actually, it seems like maybe something even stronger (like the following) would be needed:

> ## font-size dependency cycles
[...]

PR #777, opened 2018-07-04
[css-properties-values-api] Describe font-relative unit cycles.

andruud:
See #315.


andruud:
Hmm,  I guess `transform-*` can also contain font-relative units.

Issue #290, opened 2016-09-08
[css-properties-values-api] What does Typed OM return for <transform-function> type properties?


tim-loh:
For properties registered as type &lt;transform-function>, what should the Typed OM return? For the 'transform' property, the individual functions correspond to CSSTransformComponent objects, but this is not currently a CSSStyleValue.

shans:
I'm not sure it's useful to be able to register properties as being of type &lt;transform-function>. Maybe we should only allow &lt;transform-function>+?

tabatkins:
Agree that it's not useful to be a single `<transform-function>`. We should only allow registering `<transform-list>`.

wilddamon:
@shans @tabatkins I'm not sure how much design this needs, but this popped up again because a guy at Samsung decided to take a crack at implementing <transform-function>. So I've tried a very simplistic change here: https://github.com/w3c/css-houdini-drafts/compare/master...wilddamon:master

WDYT?

tabatkins:
Looks good, I went ahead and merged it. Thanks!

jyc:
Hi @wilddamon! I'm at Mozilla also trying to implement this in Servo (issue https://github.com/servo/servo/issues/17155)--do you know where the Samsung engineer is trying to implement it?

wilddamon:
Hi @jyc! Here's the CL I mentioned (although it's WIP still, since it brought up this issue again): https://codereview.chromium.org/2937043002/

jyc:
Hi @wilddamon, thanks so much! Will follow that!

tabatkins:
As @andruud pointed out in #782, we've re-allowed `<transform-function>` in the grammar.

I think the obvious answer here is that CSSTransformComponent should be a CSSStyleValue subclass, and I'm not sure why it wasn't originally.

PR #782, opened 2018-07-11
[css-properties-values-api] Clarify computed value of <transform-function>

andruud:
The current text doesn't actually mention what should happen with
`<transform-function>` values that do NOT appear in `<transform-list>`.


andruud:
OK, I just found f9dd0c0c05eff2e6efc7aacdb545cd03ff4a29ca.

@shans Was `<transform-function>` intentionally added back to the spec (e7b8c0fc46e778d17fa449344264531fab24145b) ? If so we have to reopen #290.

Issue #778, opened 2018-07-05
[css-properties-values-api] Syntax-checking in CSSStyleDeclaration.setProperty


andruud:
In the current implementation of css-properties-values-api/cssom in Blink, it is not allowed to set a registered custom property to an invalid value using setProperty. The value is just discarded.

**Should setProperty still discard invalid values for registered custom properties?**

Example:

```
<!DOCTYPE html>
<script>
    CSS.registerProperty({
        name: '--color',
        syntax: '<color>',
        initialValue: 'green',
        inherits: false
    });
</script>
<style>div { --color: red; }</style>
<div id=target></div>
<script>
    // Has no effect:
    target.style.setProperty('--color', 'url("not-a-url")');

    // Actual: red
    // Expected (by andruud): green
    console.log(getComputedStyle(target).getPropertyValue('--color'));
</script>
```

This was allegedly [1] the agreed-upon behavior at some point, but as of https://github.com/w3c/css-houdini-drafts/commit/8dfc2d985618a23f97e05c42f7ef6b083ef3256d , this behavior seems inconsistent. I think it's more in line with the new spec wording to let setProperty ignore whether the property is registered or not, and do syntax-checking computed-value time.

[1] https://codereview.chromium.org/2607403002

andruud:
The same goes for Typed OM, registering a property should have no effect on declared/inline style, but for computedStyleMap, getters return typed values after registering.

andruud:
.. or maybe not.

I guess it's kind of weird if, for a `<length>`-registered property `--foo`, this was invalid:

```
element.attributeStyleMap.set('--foo', CSS.px(20));
```

tabatkins:
The behavior is inconsistent, but intentional.

Registered custom properties don't reject invalid values at parse time *from a style sheet*, so that browsers don't have to reparse all the styles on the page to determine the correct cascade (or, equivalently, preserve the entire list of cascaded values for every custom property); instead, you just decide at computed-value time whether the value that won the cascade is valid or not.

But there's no similar implementation difficulty for checking the validity of a style set from JS.  Throwing an error gives useful feedback to the author, even if it does mean that there's now an order-dependence in the results.

(That is, setting a property then registering it gives no error, even if the value doesn't match the grammar; it just ends up as invalid at computed-value time. But registering the property and then setting it will throw an error if it doesn't match the grammar.)

I'll amend the note to make it clearer that this is an intentional inconsistency, however.

andruud:
[...]

PR #783, opened 2018-07-11
[css-typed-om] Define reification of registered custom properties.

andruud:
I don't think we need to define the reification in detail in the
[css-properties-values-api] spec. CSS Typed OM § 5 already describes
how to reify all the types mentioned in [css-properties-values-api]. I.e.
the two specs are already adequately connected though the types.

### DIFF
--- a/css-typed-om/Overview.bs
+++ b/css-typed-om/Overview.bs
@@ -2905,7 +2905,16 @@ regardless of what property it is for.
     and return the result.
 
 : registered [=custom properties=] [[css-properties-values-api-1]]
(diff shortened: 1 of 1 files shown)


andruud:
The specified values part is not completely clear yet. See #778.

tabatkins:
I don't like this. When possible, I think we should reify the specified value to the given grammar.

andruud:
@tabatkins Agreed! Updated PR.

My thinking is still that we don't need grand enumerations and mappings between syntax/types/reifications here, but I don't know what you think ...

PR #784, opened 2018-07-21
[css-layout-api][css-paint-api][css-properties-values-api] Use partial namespace CSS

foolip:
CSSOM and other CSS specs were updated in https://github.com/w3c/csswg-drafts/pull/437.

Fixes https://www.w3.org/Bugs/Public/show_bug.cgi?id=29623.


foolip:
Tests will propagate to wpt in future PRs like these:
https://github.com/web-platform-tests/wpt/pull/11819
https://github.com/web-platform-tests/wpt/pull/11893

Issue #846, opened 2018-12-12
[css-properties-values-api] @property in shadow trees


andruud:
What happens when `@property` rules appear inside shadow trees?

 * Ignored?
 * Applied (i.e. causes property registration) at the document level?
 * Some new concept of scoped/local variables?

tabatkins:
`CSS.registerProperty()` is allowed inside of shadows, and produces a global registration, which suggests that we should allow @property, and have it similarly produce a global registration (with all the clashing pain that implies).

That said, it would require us to define a stylesheet ordering of shadow stylesheets relative to the rest of the page. That seems complicated; I'm pretty sure we've avoided it so far?

I'm inclined to just disallow it, at least for now. If you're using shadow DOM, you're running script, right now; you can just call `CSS.registerProperty()`.

andruud:
> I'm inclined to just disallow it, at least for now. If you're using shadow DOM, you're running script, right now; you can just call CSS.registerProperty().

Sounds good.

Long term it should probably do something, and I'd expect e.g. `@keyframes` and `@property` to have similar semantics. But it's probably smart to tread carefully here, since `@keyframes` is a mess right now (in Chrome anyway). So I support specifying that's it's disallowed, perhaps with a note that it might be allowed in the future. It should mean `@property` isn't blocked by anything difficult and can move forward relatively quickly. 🙂

Issue #845, opened 2018-12-12
[css-properties-values-api] @property vs registerProperty


andruud:
I assume `@property` rules overwrite previous rules with the same name, e.g. if you specify the following, then `--x` is a `<length>`, and not a `<color>`:

```
@property --x { syntax: "<color>"; inherits: false; initial-value: red; }
@property --x { syntax: "<length>"; inherits: false; initial-value: 10px; }
```

However, what happens when `registerProperty` and `@property` are competing? Currently, `registerProperty` fails if something is already registered, yet for `@property` it's probably more natural that it re-registers the property.

Proposal:

 * `registerProperty` fails is the property is already registered, regardless of the origin of that registration.
 * `@property` will re-register the property if the current registration originated from `@property`, but fails (i.e. has no effect) if the current registration originated from `registerProperty`.

emilio:
Copy pasting from #902:

>  I think the easiest solution to that problem is saying that they're independent (that is, you can't unregisterProperty a property registered via @property, and vice-versa, and then you just need to specify in which order do we look at the two sets if there's a name conflict).

tabatkins:
Agree. I'm inclined to have script-registered properties win over rule-registered ones.

tabatkins:
All right, specced that @property rules for the same name have last-wins semantics, and that registerProperty() wins over @property rules.

PR #847, opened 2018-12-12
[css-properties-values-api] Add @property.

### LINKED ISSUE #137: [css-properties-values-api] Allow custom property descriptors with a CSS @-rule
The incredible flexibility of CSS Variables / Custom Properties introduces limitations compared to regular properties. By default they:
- cannot be interpolated in animations & transitions
- disable any parser-based type checking when used as a `var()`
- cannot have a default value, fallbacks must be specified each time a `var()` is used

The [CSS Properties Values API](https://drafts.css-houdini.org/css-properties-values-api/) addresses all these limitations, but requires script to run to do so.  This will cause async issues (CSS having to be re-parsed after script runs) and is no use at all in cases where scripts are disabled for security reasons (e.g., SVG-as-image).

While the general argument for using JS APIs for Houdini has been discussed in #86, I think property declarations are a special case.  This isn't about declaring functionality that needs to be represented as script functions; it's merely declaring a data object.  An @-rule of descriptors and values seems a perfect fit:

``` css
@property --highlight-color {
  syntax: "<color>";
  initial-value: red;
  inherits: true;
}

@property --gap-spacing {
  syntax: "<length-percentage>";
  initial-value: 1em;
  inherits: false;
}
```

It would of course be preferable to always have `@property` rules earlier in the CSS parsing sequence than any declarations using those properties.  However, since it is also going to be possible to set property declarations asynchronously with the API, I don't think this would need to be an enforced syntax rule, just a performance recommendation.

The effects of this declaration would be the same as for a declaration via the API.  E.g., If `initial-value` isn't specified, it would get the "nothing" initial value described in [CSS Custom Properties](https://drafts.csswg.org/css-variables/#defining-variables), with the resulting implication for `var()` function fallbacks.  

The API currently advises throwing an error if the provided initial value doesn't match the provided syntax.  For a `@property` declaration, I would recommended simply dropping the non-conforming initial value declaration.  This could allow multiple declarations with standard CSS fallback methods, e.g., for new functions that might not be recognized:

``` css
@property --highlight-color {
  syntax: "<color>";
  initial-value: red;
  initial-value: lighten(maroon);
  inherits: true;
}
```

andruud:
This adds basic support for `@property`, without solving any of the potentially complicated issues, like #845 and #846.

Resolves #137, at least partially.

### DIFF
--- a/css-properties-values-api/Overview.bs
+++ b/css-properties-values-api/Overview.bs
@@ -914,16 +914,17 @@ and affects how the custom property calculates its [=computed value=].
 The <dfn>@property</dfn> rule {#at-property-rule}
 =================================================
 
-The ''@property'' rule provides an alternative way to register a custom
-property. Valid ''@property'' rules result in a
-[=register a custom property|registered custom property=], as if
-{{registerProperty()}} had been called with equivalent parameters.
+The ''@property'' rule provides an alternative way to register a custom property,
+directly in a stylesheet
+without having to run any JS.
+Valid ''@property'' rules result in a [=register a custom property|registered custom property=],
+as if {{registerProperty()}} had been called with equivalent parameters.
 
-The semantics of registered properties are the same regardless of
-the mechanism used to perform the registration. This means that, once
-registered, it does not matter whether the registration originated from
-{{registerProperty()}} or ''@property'': the property has the same
-behavior either way.
+The semantics of registered properties are the same
+regardless of the mechanism used to perform the registration.
+This means that, once registered,
+it does not matter whether the registration originated from {{registerProperty()}} or ''@property'':
+the property has the same behavior either way.
 
 The syntax of ''@property'' is:
 
@@ -933,21 +934,22 @@ The syntax of ''@property'' is:
 	}
 	</pre>
 
-''@property'' rules require a 'syntax', and 'inherits' descriptor; if either are
-missing, the ''@property'' rule is invalid, and must not result in a registration.
-The 'initial-value' descriptor is optional only if the syntax is the
-[=universal syntax descriptor=], otherwise the descriptor is required; if it's
-missing, the ''@property'' rule is invalid,
-and must not result in a registration.
+''@property'' rules require a 'syntax' and 'inherits' descriptor;
+if either are missing,
+the entire rule is invalid and must be ignored.
+The 'initial-value' descriptor is optional
+only if the syntax is the [=universal syntax descriptor=],
+otherwise the descriptor is required;
+if it's missing, the entire rule is invalid and must be ignored.
 
-The ''@property'' rule accepts the descriptors defined in this chapter.
-Declarations that contain other descriptors not defined here, must be ignored to
-maintain forwards compatiblity.
+If an ''@property'' rule contains any unknown descriptors,
+the entire rule is invalid and must be ignored.
 
 Issue(845): What happens when we have both an ''@property'' rule, and a
 {{registerProperty()}} call for the same property?
 
 Issue(846): What happens when ''@property'' appears inside a shadow tree?
+
 
 The 'syntax' descriptor {#the-syntax-descriptor}
 ------------------------------------------------
@@ -956,18 +958,19 @@ The 'syntax' descriptor {#the-syntax-descriptor}
 	Name: syntax
 	Value: <<string>>
 	For: @property
-	Initial: "*"
+	Initial: n/a (see prose)
 	</pre>
 
-Specifies the syntax of the custom property, on the form defined by
-[[#syntax-strings]]. This descriptor is equivalent to the
-{{PropertyDescriptor/syntax|syntax}} member of {{PropertyDescriptor}}.
+Specifies the syntax of the custom property,
+in the form defined by [[#syntax-strings]].
+This descriptor is equivalent to the {{PropertyDescriptor/syntax|syntax}} member of {{PropertyDescriptor}}.
 
-The syntax descriptor is required for the ''@property'' rule to be valid;
-if it's missing, the ''@property'' rule has no effect.
+The 'syntax' descriptor is required for the ''@property'' rule to be valid;
+if it's missing, the ''@property'' rule is invalid.
 
-If the provided string does not successfully parse as a syntax descriptor,
-the ''@property'' rule has no effect.
+If the provided string does not successfully [=consume a syntax definition|parse as a syntax definition=],
+the descriptor is invalid and must be ignored.
+
 
 The 'inherits' descriptor {#inherits-descriptor}
 ------------------------------------------------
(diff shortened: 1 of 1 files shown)


heycam:
@tabatkins @astearns @therealglazou could one of you review and merge this PR if it looks good?  Since @andruud is beginning implementation it would be preferable to have this in the spec, so that others can look at it...

css-properties-values-api/Overview.bs
```diff
@@ -601,6 +601,125 @@ referenced, otherwise the declaration is
...
+===========================================================================
+
+The ''@property'' rule provides an alternative way to register a custom
+property. Well-formed ''@property'' rules result in a
+[=register a custom property|registered custom property=], as if
+{{registerProperty()}} had been called with equivalent parameters.
+
+The semantics of registered properties are the same regardless of
+the mechanism used to perform the registration. This means that, once
+registered, it does not matter whether the registation originated from
```
astearns:
```suggestion
registered, it does not matter whether the registration originated from
```

css-properties-values-api/Overview.bs
```diff
@@ -601,6 +601,125 @@ referenced, otherwise the declaration is
...
+
+This descriptor may be omitted when the value of the 'syntax' descriptor is
+<code>"*"</code>, in which case the initial value is
+[[css-variables#invalid-variables|the empty value invalid for substituion]]
+described by [[css-variables]]. Otherwise, the 'initial-value' descriptor is
+required, and the ''@property'' rule will not register the property if it's
+omitted.
+
+The value specified for 'initial-value' must be [=computationally independent=],
+and match the grammar corresponding to the specified 'syntax'.
```
astearns: What happens if it is not one or the other of these things?

css-properties-values-api/Overview.bs
```diff
@@ -601,6 +601,125 @@ referenced, otherwise the declaration is
...
+The <code>CSSPropertyRule</code> interface {#the-css-property-rule-interface}
+-----------------------------------------------------------------------------
+
+The <code>CSSPropertyRule</code> interface represents an ''@property'' rule.
+
+<pre class='idl' export>
+[Exposed=Window]
+interface CSSPropertyRule : CSSRule {
+    attribute CSSOMString name;
+    readonly attribute CSSStyleDeclaration style;
```
astearns: why not have the descriptors as bare attributes here?

astearns:
I took a look, but @tabatkins should review.

emilio:
What happens when you dynamically remove an `@property` rule? What about multiple `@property` rules with the same name?

css-properties-values-api/Overview.bs
emilio: +1, CSSStyleDeclaration has a couple interop problems (WebKit and blink define all CSS properties within it, so you can have `'width' in fontFaceRule.style` return true, which is broken IMO) which I'd rather not repeat.

css-properties-values-api/Overview.bs
andruud: Ah, yes good point.

(The whole rule will be invalid).

css-properties-values-api/Overview.bs
andruud: :+1: Will do.

andruud:
Thank you @heycam, @astearns & @emilio.

> What happens when you dynamically remove an `@property` rule?

Then I'd say the property isn't registered anymore. There is a note saying we might add a way to unregister a property ... I guess this does that.

> What about multiple `@property` rules with the same name?

Yeah, this is interesting (since CSS.registerProperty throws if you try to re-register). I assume we want the last one to apply for `@property`, even if that's the opposite behavior of CSS.registerProperty). WDYT?

emilio:
Seems ok, though kind of unfortunate... In general having two ways of doing the same is not great IMHO. But "last wins" seems consistent with `@keyframes` so it seems fair.

emilio:
Also, I'd note that with the decision we took in #880, it means that `setProperty` and `CSS.supports` may need to update style data (I think the Blink concept is doing an active style update?).

I think that's terrible, fwiw, and if I'd have known there were plans to introduce a declarative way of defining these I'd have objected even more strongly to that.

andruud:
@emilio Please excuse my ignorance, but why would setProperty/CSS.supports need to update style data? (And what is "style data" anyway?)

andruud:
@emilio Oh, never mind, I see what you mean ... hmm ...

emilio:
Yeah, sorry, style data is a Servo-ism, it means the "processed" data out of the parsed stylesheets. This is why I used the "active style" concept which is a WebKit-ism :)

But yeah, it basically means that you need to look at whether the processed data from stylesheets is dirty and update it from a `setProperty` call (since there may be new `@property` rules for example that you still haven't processed), which I think is really bad, not only for the unexpected, potentially big amount of work that that requires, but also because `setProperty` at least is a way to potentially dirty this data. Of course you can try to be smarter about invalidating and updating it, but I think it's not great even with that.

css-properties-values-api/Overview.bs
```diff
@@ -911,6 +911,160 @@ and affects how the custom property calculates its [=computed value=].
...
+Issue(846): What happens when ''@property'' appears inside a shadow tree?
+
+The 'syntax' descriptor {#the-syntax-descriptor}
+------------------------------------------------
+
+	<pre class='descdef'>
+	Name: syntax
+	Value: <<string>>
+	For: @property
+	Initial: "*"
```
lahmatiy: `*` is wrong syntax, since means `zero or more occurrences` multiplier with missed term before it. Looks like it should be `Nothing` like for other descriptors if it's a required property, or `"<declaration-value>"` if it will become optional.

andruud: It's not `*`, it's `"*"`.

lahmatiy: Yeah, I mean wrong syntax for a custom property value (inside quotes used specific grammar). So, `*` is not valid for that

andruud: This accepts a `<string>`, hence `"*"` is valid in that sense.

What is or isn't a valid _syntax descriptor_ is determined by [3.4.2. Consume a syntax descriptor](https://drafts.css-houdini.org/css-properties-values-api-1/#consume-syntax-descriptor), not css3-values (directly).

In other words, a single `*` alone is a special snowflake.

lahmatiy: I see, thank you for clarification!

css-properties-values-api/Overview.bs
```diff
@@ -933,41 +934,43 @@ The syntax of ''@property'' is:
...
+The 'initial-value' descriptor is optional
+only if the syntax is the [=universal syntax descriptor=],
+otherwise the descriptor is required;
+if it's missing, the entire rule is invalid and must be ignored.
 
-The ''@property'' rule accepts the descriptors defined in this chapter.
-Declarations that contain other descriptors not defined here, must be ignored to
-maintain forwards compatiblity.
+If an ''@property'' rule contains any unknown descriptors,
+the entire rule is invalid and must be ignored.
```
andruud: @tabatkins Ouch, my forwards compat! :scream:

tabatkins: Oh, I misread what you'd wrote there!

PR #853, opened 2019-01-21
[css-typed-om] Use `partial interface mixin ElementCSSInlineStyle`

foolip:
https://drafts.csswg.org/cssom/#the-elementcssinlinestyle-interface is an
interface mixin, and the partial must match this. Update the prose to reflect
this as well.


foolip:
@tabatkins can you review this?

foolip:
Ping @tabatkins

PR #858, opened 2019-02-11
[css-layout-api] Name the FragmentResultOptions argument

foolip:
This was invalid Web IDL. Also make it optional since there are no required members.


foolip:
@bfgeek can you review this?

PR #860, opened 2019-02-15
[css-properties-values-api] Define parsing rules for the syntax string.

### LINKED ISSUE #112: need to define the exact grammar accepted by "syntax"
We need to define the grammar for the string used for describing the syntax of CSS properties.  I'm not sure that using CSS value metasyntax is appropriate for this, or at the very least would be a little confusing. :-)

A question that would answer, for example, is whether `" *"` would be accepted, or what exact white space characters are accepted around the bar in `"<length> | <number>"`.

andruud:
This defines how to parse the syntax strings using some algorithms and
concepts from css-syntax, while leaning as little as possible on its
concept of tokens.

* Whitespace behavior is now defined.
* Escaping is now allowed for ident syntax strings.

Resolves #112.

### DIFF
--- a/css-properties-values-api/Overview.bs
+++ b/css-properties-values-api/Overview.bs
@@ -140,25 +140,11 @@ as arguments of the same names.
 		<a>throw</a> an {{InvalidModificationError}}
 		and exit this algorithm.
 
-	3. If |syntax| is not present,
-		or is equal to <code>"*"</code> (U+002A ASTERISK),
-		let |parsed syntax| be undefined,
-		and skip to the next step of this algorithm.
+	3. Attempt to parse |syntax| according to the rules in [[#parsing-the-syntax-string]].
+		If it does not parse successfully, <a>throw</a> a {{SyntaxError}}.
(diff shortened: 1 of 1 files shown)


tabatkins:
I added some fixups. Still needs some more work, as various things aren't linking, but it's usable right now.

One point tho: CSS allows `+#` as a grammar multiplier; that is, multipliers can be stacked. Is there a good reason not to allow that?

andruud:
Re. stackable multipliers, I don't think anything prevents us from doing that. I was only maintaining limitations that were already in the spec. (According to my interpretation, that is). The previous spec text said something like "`+`/`#` multiplies any _primitive term_".

I don't think we should scope-creep this spec any further right now, in any case. There are still too many gaps on the existing functionality, e.g. CSS(Typed)OM behavior, and substitution behavior. Apparently the cascading behavior / syntax-checked-at-computed-value-time concepts were also not clear to TAG.

tabatkins:
Sounds reasonable, I'll bring it up as an improvement later.

PR #861, opened 2019-02-18
[css-properties-values-api] "Consume a name" returns a string.

andruud:
"Consume a name" doesn't return an `<ident-token>`. Also, it seems more
clean to avoid css-syntax's tokens if we can.


tabatkins:
Ah shoot, I thought it returned an ident token, so I was then grabbing the ident token's value (a string) to continue the algorithm with. ^_^ Def my bad.

PR #866, opened 2019-03-07
[css-animationworklet] Change state to be a function instead of getter

majido:
There is no need for state to be a getter rather than a function.

It is simpler to implement and more consistent with `animate()` for state to also be a function.


css-animationworklet/Overview.bs
```diff
@@ -253,21 +253,21 @@ effect is mutable only in a certain thread). Animation worklet guarantees that a
 instance's state is maintained even if the instance is respawned in a different global scope.
 
 The basic mechanism for maintaining the state is that the animation worklet snapshots the local
-state that is exposed via {{StatefulAnimator/state}} attribute and then reifies it at a later time
-to be passed into the constructor when the animator instance is respawned in a potentially different
-global scope. This processes is specified is details in <a>migrate an animator instance</a>
-algorithm.
+state that is exposed via {{StatefulAnimator/state}} function and then reifies it so that it can be
```
stephenmcgruer: Nit; 'via [the] {{StatefulAnimator/state}} function'?

PR #873, opened 2019-04-04
[css-properties-values-api] Define effects on CSSOM.

andruud:
* Define how setProperty() behaves for registered custom properties.
 * Define how values reify.
 * How underlying values are created is not defined, as I _think_ the
   current description in css-typed-om covers it.

### DIFF
--- a/css-properties-values-api/Overview.bs
+++ b/css-properties-values-api/Overview.bs
@@ -819,12 +819,20 @@ This means that once a property is registered, it is not possible to use
 {{setProperty()}} with a value that violates the registered syntax of the
 property.
 
-However, note that registering a property '--x' does not invalidate
-already-specified values for '--x', even if those values violate the
-registered syntax.
+Note: Registering a property with a particular syntax
+does not invalidate already-specified values for the property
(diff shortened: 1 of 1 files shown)


andruud:
This is the final missing "substantial" piece of this spec, I think.

Not sure if it's OK to use `<declaration-value>` in this way. CSS Typed OM seems to have a concept of underlying values "being" a certain type (rather than e.g. matching a production), in which case it should be fine.

tabatkins:
Did some minor tweaking to wording and linking, but nothing to the algos. Looks good otherwise, sorry for the delay in reviewing.

> CSS Typed OM seems to have a concept of underlying values "being" a certain type (rather than e.g. matching a production), in which case it should be fine.

Yeah, underlying values are an implementation-defined structure, that just must be able to reproduce the necessary data for reification/stringification. What you have down is fine; the only thing that's not a `<declaration-value>` but still a valid custom property value is the empty value, and that can't ever match a syntax, so it's fine to let that fall into the second case.

andruud:
Thanks Tab :+1:

PR #884, opened 2019-05-06
[css-properties-values-api] Require <custom-ident> as component name.

### LINKED ISSUE #879: [css-properties-values-api] Consume a component name presumably shouldn't allow CSS wide keywords, but it does.
https://drafts.css-houdini.org/css-properties-values-api-1/#consume-syntax-component:

When you hit a `name-start code point`:

> If the stream starts with an identifier, reconsume the current input code point from stream then consume a name from stream, and set component’s name to the returned `<ident-token>`’s value. Otherwise return failure.

Per https://github.com/w3c/css-houdini-drafts/issues/259, I assume that it should match `<custom-ident>`. Blink does throw for this:

```js
CSS.registerProperty({ name: "--bar", syntax: "inherit", inherits: false, })
```

cc @andruud @tabatkins

andruud:
Something like this was in the original syntax parsing PR, but it got
lost at some point.

Resolves #879.

### DIFF
--- a/css-properties-values-api/Overview.bs
+++ b/css-properties-values-api/Overview.bs
@@ -640,11 +640,11 @@ Parsing the syntax string {#parsing-syntax}
 		and set |component|’s |name| to the returned value.
 		Otherwise return failure.
 
+		If |component|’s |name| does not match the <<custom-ident>> production,
+		return failure.
+
 	:   anything else
 	::  Return failure.
-
-	If |component|’s |name| does not match the <<custom-ident>> production,
-	return failure.
 
 	If |component|’s |name| is a [=pre-multiplied data type name=],
 	return |component|.
--- a/css-properties-values-api/Overview.bs
+++ b/css-properties-values-api/Overview.bs
@@ -643,6 +643,9 @@ Parsing the syntax string {#parsing-syntax}
 	:   anything else
 	::  Return failure.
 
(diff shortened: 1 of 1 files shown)


andruud:
cc @tabatkins @emilio

andruud:
Uhm, oops, that doesn't work, hold on.

andruud:
There we go.

css-properties-values-api/Overview.bs
```diff
@@ -640,6 +640,9 @@ Parsing the syntax string {#parsing-syntax}
 		and set |component|’s |name| to the returned value.
 		Otherwise return failure.
 
+		If |component|’s |name| does not match the <<custom-ident>> production,
```
tabatkins:
```suggestion
		If |component|’s |name| does not [=CSS/parse=] as a <<custom-ident>>,
```

andruud: Yeah, that makes sense. Thanks. 👍

andruud: That means the following probably isn't great either:

> Any sequence which starts an identifier, can be consumed as a name, and matches the &lt;custom-ident> production

tabatkins: Ah, indeed.

PR #885, opened 2019-05-06
[css-paint-api] Lean on "syntax definition" concept.

andruud:
The css-properties-values-api now describes in detail how to produce a
"syntax descriptor" from a string, so it's better to link to that rather
than linking to a section in the spec.

### DIFF
--- a/css-paint-api/Overview.bs
+++ b/css-paint-api/Overview.bs
@@ -142,7 +142,7 @@ by the paint function). It consists of:
     <code>DOMStrings</code>.
 
  - A <dfn for="document paint definition">input argument syntaxes</dfn> which is a [=list=] of
-    parsed [[css-properties-values-api-1#supported-syntax-strings]].
+    [=syntax descriptors=].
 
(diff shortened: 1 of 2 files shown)


andruud:
@bfgeek ptal

Is it possible to get newly "dfn export"-ed things to link locally? I can't get it to work ...

tabatkins:
> Is it possible to get newly "dfn export"-ed things to link locally? I can't get it to work ...

You need to wait for Shepherd to make a pass over the defining spec (runs on each push, but can be delayed by other jobs, so 30-60 minutes is reasonable), then make sure you run `bikeshed update` locally to get the new definitions.

andruud:
@tabatkins OK, in other words, this needs to be merged to master before I can `bikeshed update`, right?

tabatkins:
The properties-and-values definition needs to be in master before paint-api can use it, yeah.

(You can just commit the link now and it'll automatically reparse into a valid link later when things sync up.)

Issue #880, opened 2019-04-29
[css-properties-values-api] Interaction with setProperty is unfortunate.


emilio:
The reasoning for that is intentional apparently:

  https://github.com/w3c/css-houdini-drafts/issues/778#issuecomment-406440820

But I think it's still weird that:

```js
let div = document.createElement("div");
div.style.setProperty("--foo", "bar");
```

Behaves differently to:

```js
let div = document.createElement("div");
div.setAttribute("style", "--foo: bar");
```

Is this behavior up-for-discussion? I think it is really unfortunate.

cc @andruud @tabatkins

emilio:
If `setProperty` behaves that way, `CSS.supports` two-value function should probably behave that way. But I still think it'd be much consistent and better over-all if custom properties would only be syntax-checked at computed-value time.

tabatkins:
At bare minimum, we want TypedOM to do custom-property parsing; it's a good spot to give feedback to the script author that they've made a mistake.

I'd *ideally* have liked to do the whole-world reparse; I don't like the inconsistency either, and it limits the power of registered custom properties in a troublesome way, since now your assumed-under-the-script's-control custom property is beholden to what CSS thinks is valid, but without the protection of fallback. But oh well.

I'd be *willing* to shift on the middle ground of string-OM setProperty(); all I want is some degree of meaningful consistency, so authors can confidently predict when each behavior will occur. So there's a few different places I can see us drawing the error/no-error line:

* "never invalidate from any source": means that explicitly using TypedOM constructed values can't even give feedback to the author, which kinda sucks imo.
* "Typed OM" vs "everything else": This kind of siloing based on specification-origin is awkward; Typed OM does string parses (from `.set()` and `.parse()`), so it's a little more difficult for authors to predict what'll check/throw and what won't.
* "object-based value APIs" vs "string-based APIs": TypedOM's `.set()` and `.parse()` wouldn't throw for bad string values, but passing objects to `.set()` would. Consistent in usage, and so at least somewhat predictable/learnable by authors, but also quite limited.
* "parses caused by HTML" vs "parses caused by JS": stylesheets from `link` and `style` elements, or `style` attributes, don't check validity (whether they were in the static markup or added/modified later via DOM), but any CSSOM operations or other JS things do (including things like Constructable Stylesheets, both forms of CSS.supports(), `el.style["--foo"]` if we end up speccing that, etc.)
[...]

PR #889, opened 2019-05-10
[css-properties-values-api] Describe behavior of CSS.supports(property, value).

andruud:
The Note is a little fuzzy, at best ... and it's not obvious exactly
where it's best to put it. It might make sense in the conditional-rules
section, since it explains the preceding CSS.supports(conditionText) vs
CSS.supports(property, value) situation.

Resolves issue #880.


andruud:
@tabatkins, please have a look. Thanks <3

tabatkins:
(fixed in a different way, so closing. thanks!)

andruud:
@tabatkins :+1:

Issue #834, opened 2018-11-22
[css-animationworklet] Listening events in animation worklet


LordGuard:
Can I listen events in animation worklet or transfer data from main thread to running animation worklet?

majido:
## On transferring data from main thread

I think this the need such a mechanism to make animation worklet useful. Since many useful effects running inside animation worklet may need to know about specific values from main thread, e.g., size of window or bounding client rects of various objects.

Since we already allow data to be sent when worklet animation (via `options`) when it is constructed. The logical extension is to allow options to be updated as well. This is a simple and reasonable extension. I will update the spec to allow this.

See  more details on specific proposed change here: https://bugs.chromium.org/p/chromium/issues/detail?id=932619


## On listening to events in animation worklet

Again I think there is a lot of merit to this idea. In fact, one of [our goals](https://github.com/w3c/css-houdini-drafts/blob/master/css-animationworklet/principles.md#third-principle---interactivity) is to allow worklet animation inside animation worklet context to be able to react to user input (touch, gesture, etc.)

We have been trying to come up with the right abstraction and API to allow this (see: [1](https://docs.google.com/document/d/1-APjTs9fn4-E7pFeFSfiWV8tYitO84VpmKuNpE-25Qk/edit), [2](https://github.com/w3c/csswg-drafts/issues/2493), [3](https://github.com/w3c/csswg-drafts/issues/2493#issuecomment-422109535)), and our recent idea which I am very hopeful about is captured [here](https://docs.google.com/document/d/1byDy6IZqvaci-FQoiRzkeAmTSVCyMF5UuaSeGJRHpJk/edit#heading=h.dc7o68szgx2r). 


I think once we have the solution for sending data from main thread to worklet, one can build a simple event forwarding on top of it. But we need a better solution to have low-latency event forwarding which the other ideas are trying to address.

majido:
I filed a [separate issue](https://github.com/w3c/css-houdini-drafts/issues/869) for sending data from main to worklet since I believe the usecases and APIs differ. Lets keep this issue focused on receiving events inside worklet.

majido:
The proposal for allowing input events to worklets is now published under WICG: https://discourse.wicg.io/t/proposal-exposing-input-events-to-worker-threads/3479

Here is the [explainer](https://github.com/NavidZ/input-for-workers) for it. I am working closely with other folks on this to make sure it addresses Animation Worklet usecases.

PR #890, opened 2019-05-16
[css-animationworklet] Remove multiple timelines

majido:
Drop multiple timelines and relevant concepts for Animation Worklet.

This also drops the hide-bar example from spec. I may re-work this example (e.g., switching timeline via setting animation.timeline on main thread)  and add it back this in future but for now dropping it is simpler.

Multiple timeline was originally the solution that would allow seamless switching between between
scroll timeline and regular timelines. Since then we have iterated over multiple different solutions.
See #834 for a list of various design we have considered. It seems that allowing events to be handled
in Animation Worklets will enable the relevant usecases. 

For now we drop multiple timelines fromAnimation Worklet Level-1 to ensure simplicity and better alignment with WebAnimations.

### DIFF
--- a/css-animationworklet/Overview.bs
+++ b/css-animationworklet/Overview.bs
@@ -379,7 +379,7 @@ Animator Instance {#animator-instance-section}
 
 An <dfn>animator instance</dfn> is a <a>struct</a> which describes a fully realized custom animation
 instance in an {{AnimationWorkletGlobalScope}}. It has a reference to an <a>animator definition</a>
-and owns the instance specific state such as animation effect and timelines. It consists of:
+and owns the instance specific state such as animation effect and timeline. It consists of:
 
  - An <a>animator name</a>.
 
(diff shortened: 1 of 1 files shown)


css-animationworklet/Overview.bs
```diff
@@ -690,7 +686,7 @@ Here are a few implications of the above semantics:
     parallel execution context.
 
 Issue(811): Come with appropriate mechanism's for <a>animator instance</a> to get notified when its
-   animation currentTime is changing e.g., via reverse(), finish() or playbackRate change. So that
+   animation curren is changing e.g., via reverse(), finish() or playbackRate change. So that
```
flackr: Typo?

css-animationworklet/Overview.bs
```diff
@@ -848,96 +843,8 @@ There are no known privacy issues introduced by these features.
 Examples {#examples}
 ====================
 
-Example 1: Hidey Bar. {#example-1}
```
flackr: With the removal of this should we call out that this is one use case which isn't currently possible?

majido: This seems like a job for EXPLAINER which I have partially updated. I plan to add the same example but using "input for worklet" instead.

PR #894, opened 2019-05-29
[css-properties-values-api] Fix syntax string parsing to not treat whitespace between components the same as pipes.

### LINKED ISSUE #893: [css-properties-values-api] The fact that "foo bar" and "foo | bar" are equivalent syntax is confusing.
In https://drafts.css-houdini.org/css-properties-values-api-1/#consume-syntax-descriptor, `|` and any other character are treated basically equivalently, so it means that `<length> <percentage>` is really a syntax for `<length> | <percentage>`, rather than a `<length>`, then a `<percentage>`.

Is this intentional? Is there any reason we're closing the gate to more expressive syntaxes that allow multiple keywords?

emilio:
Fixes #893

### DIFF
--- a/css-properties-values-api/Overview.bs
+++ b/css-properties-values-api/Overview.bs
@@ -586,33 +586,24 @@ Parsing the syntax string {#parsing-syntax}
 		preprocessed as specified in [[css-syntax-3]].
 		Let |descriptor| be an initially empty [=list=] of <a>syntax components</a>.
 
-	5. Repeatedly consume the <a>next input code point</a> from |stream|:
+	5. <a>Consume a syntax component</a> from |stream|.
+		If failure was returned,
+		return failure;
+		otherwise,
+		[=list/append=] the returned value to |descriptor|.
 
-		:   <a>whitespace</a>
-		::  Do nothing.
+		Consume as much <a>whitespace</a> as possible from |stream|.
 
+		Consume the <a>next input code point</a> in |stream|:
 		:   EOF
-		:: If |descriptor|’s [=list/size=] is greater than zero,
-			return |descriptor|;
-			otherwise, return failure.
+		:: return |descriptor|.
 
 		:   U+007C VERTICAL LINE (|)
-		:: If |descriptor|’s [=list/size=] is greater than zero,
-			<a>consume a syntax component</a> from |stream|.
-			If failure was returned,
-			return failure;
-			otherwise,
-			[=list/append=] the returned value to |descriptor|.
+		:: Repeat step 5.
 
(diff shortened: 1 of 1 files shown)


andruud:
Oh, nice, thanks @emilio <3

I'm not actually an editor though, so I can't merge it.

--> @tabatkins 

(It would be nice to have assign-review-to-people privileges, so I didn't have to ping people in a comment, like a commoner, djeez).

css-properties-values-api/Overview.bs
```diff
@@ -596,19 +596,15 @@ Parsing the syntax string {#parsing-syntax}
...
 		:: If |descriptor|’s [=list/size=] is greater than zero,
-			<a>consume a syntax component</a> from |stream|.
-			If failure was returned,
+			and the <a>current input code point</a> is not U+007C VERTICAL LINE (|),
 			return failure;
-			otherwise,
-			[=list/append=] the returned value to |descriptor|.
 
-			If |descriptor|’s [=list/size=] is zero, return failure.
+		:: If the <a>current input code point</a> is not U+007C VERTICAL LINE (|),
```
tabatkins: These should all be paragraphs under the one dd, not multiple separate dds

emilio: Sure, done.

emilio:
I don't seem to have merge access to this repo, so if you could do that @tabatkins it'd be awesome.

Thanks!

andruud:
@emilio Just discovered that this allows a leading "|", e.g. "|foo" is a valid syntax.

andruud:
Let's rethink step 5 as a whole. Drop the "repeatedly consume" thing and instead do:

Step 5:
 * Consume a syntax component from stream. If failure was returned, return failure; otherwise, append the returned value to descriptor. [Note: Consume a syntax component consumes leading whitespace, which is relevant for repeats of this step].
 * Consume as much whitespace as possible from stream.
 * Consume the next input code point from stream:
   * EOF: Return descriptor. [Note: I don't think we still need the descriptor size check, since we'll now exit early if "Consume a syntax component" fails?]
   * U+007C VERTICAL LINE (|):  Repeat this step.
   * Anything else: Return failure.

WDYT?

emilio:
Err, yes indeed it does.

Or alternatively, also reconsume the current code-point even if it's `|`, if the output is empty. But yeah, that sounds cleaner.

emilio:
Your algorithm also needs to skip trailing whitespace after parsing a component, otherwise you stop parsing `foo | bar` unless it's written `foo|bar`

emilio:
Err, you did that already, should learn to read. Yeah, that looks good.

PR #911, opened 2019-07-02
[css-properties-values-api] HTML entities in Animation Behavior section.

andruud:


tabatkins:
Ah, no, the problem is that I was using inline Markdown shorthands without turning that on in the metadata. Fixed it by adding an appropriate metadata line; thanks for the catch.

PR #912, opened 2019-07-02
[css-properties-values-api] Generally ignore syntax in APIs.

### LINKED ISSUE #902: [css-properties-values-api] @property vs setProperty
Given an `@property` rule in some stylesheet, how should the two-param `setProperty()`/`CSS.supports()` behave? They probably need to take into account the registration caused by the `@property` rule, even if that registration has not yet been effectuated.

Updating "style data" (as @emilio [calls it in this comment](https://github.com/w3c/css-houdini-drafts/pull/847#issuecomment-502497184)) during `setProperty()` seems nasty.

Do we revisit #880? Give up on `@property`? Accept the nastiness? Something else?

cc @tabatkins

andruud:
Resolves #902.

### DIFF
--- a/css-properties-values-api/Overview.bs
+++ b/css-properties-values-api/Overview.bs
@@ -711,21 +711,10 @@ Conditional Rules {#conditional-rules}
 --------------------------------------
 
 <div class=note>
-	''@supports'' rules and the {{CSS/supports(conditionText)}} method
+	''@supports'' rules and the {{CSS/supports}} method
 	behave as specified in [[!css-variables]],
 	without distinguishing between registered and unregistered custom properties,
 	and paying no attention to registered properties' syntaxes.
-
-	{{CSS/supports(property, value)}},
-	as specified in [[css-conditional-1]],
-	<em>does</em> pay attention to the syntax of registered custom properties.
-
-	The general principle in use
-	is that anything which evaluates the validity of a CSS property value on its own
-	will pay attention to the syntax of registered custom properties,
-	but anything parsing any larger construct
-	(such as declarations, or entire stylesheets)
-	will ignore the syntax.
 </div>
 
 Relative URLs {#relative-urls}
@@ -820,57 +809,12 @@ string.
 CSSOM {#cssom}
 ==============
 
-{{CSSStyleDeclaration}} Behavior {#css-style-declaration-behavior}
-------------------------------------------------------------------
-
-The <a>set a CSS declaration</a> algorithm gains an additional constraint
-for registered custom properties:
-
-* The target declaration must have a value that matches the registered
-	[=syntax descriptor|syntax=] of the property.
-
-This means that once a property is registered, it is not possible to use
-{{setProperty()}} with a value that violates the registered syntax of the
-property.
-
-Note: Registering a property with a particular syntax
-does not invalidate already-specified values for the property
-(that is, they're not thrown out at parse-time,
-like invalid properties usually are),
-even if those would violate the registered syntax.
-The registration <em>does</em> prevent new invalid values from being set,
-and affects how the custom property calculates its [=computed value=].
-
-<div class='example'>
-	For example,
-	in the following code snippet,
-	the ''--x'' property is set to a value
-	that doesn't match the later-registered syntax:
-
-	<pre class="lang-javascript">
-		document.body.style.setProperty('--x', '10px');
-		CSS.registerProperty({
-			name: '--x',
-			syntax: '&lt;color>',
-			initialValue: 'white',
-			inherits: false
-		});
-
-		// This still outputs "10px", as existing values aren't
-		// re-evaluated after a syntax is registered.
-		console.log(document.body.style.getPropertyValue('--x'));
-
-		// This silently fails, like any other property would
-		// when one attempts to set it to an invalid value.
-		document.body.style.setProperty('--x', '20px');
-
-		// Because ''--x'' is still set to "10px", which isn't
-		// a valid <<color>>, it will be "<l>[=invalid at computed-value time=]</l>",
-		// falling back to the registered initial value of "white".
-		// 'background-color' will then receive that color and get
-		// set to "white".
-		document.body.style.backgroundColor = "var(--x)";
-	</pre>
+<div class=note>
+	The value specified for a registered custom property is not
+	interpreted until computed-value time. This means that only APIs that
+	retrieve computed values are affected. Other APIs must ignore the
+	{{[[registeredPropertySet]]}} slot of the associated {{Document}}, and
+	treat all custom properties as unregistered.
 </div>
 
 {{CSSStyleValue}} Reification {#css-style-value-reification}
@@ -880,14 +824,8 @@ and affects how the custom property calculates its [=computed value=].
 	To <dfn>reify a registered custom property value</dfn> given a property
 	|property| and [=syntax descriptor=] |syntax|, run these steps:
 
-	For specified values:
-
-		1. If the value is a <<declaration-value>>, and |syntax| is not the
-			[=universal syntax descriptor=], attempt to [=CSS/parse=]
-			the value according to |syntax|. If this succeeds, [=reify=]
-			the result as if it were a computed value, and return that result.
-		2. Otherwise, [=reify a list of component values=] from the value, and
-			return the result.
+	For specified values, [=reify a list of component values=] from the value,
+	and return the result.
 
 	For computed values:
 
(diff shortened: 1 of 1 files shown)


andruud:
:crying_cat_face: (but also :tada:)

@emilio, can you take a look at this and verify that this is what we want?

css-properties-values-api/Overview.bs
```diff
@@ -711,21 +711,10 @@ Conditional Rules {#conditional-rules}
 --------------------------------------
 
 <div class=note>
-	''@supports'' rules and the {{CSS/supports(conditionText)}} method
+	''@supports'' rules and the {{CSS/supports}} method
```
emilio: s/method/methods?

tabatkins: "method" is fine (there's only one of them! overloads are a WebIDL fiction!), but it should be `{{CSS/supports()}}`.

tabatkins:
Hmmm, losing the reification behavior sucks. :( It means there's no way to interact with, say, a `"<length>"`-registered property as a CSSUnitValue, unless you go parse it yourself with your own knowledge.

Is there any way we can save this? It'll require flushing style data and be racy with stylesheets, unfortunately.

Even if we don't think it's reasonable to salvage `.get()`'s reification behavior, I'd want *some* way to say "reify this as --foo property specifies". But anything we add will *also* be racy. We can't even rely on, say, using a promise that doesn't resolve until the property is registered, since there might be multiple registrations.

andruud:
> losing the reification behavior sucks. :( 

From the author's perspective, yeah :crying_cat_face:. But a `<length>`-registered property being `CSSUnparsedValue` at parse-time/specified-value-"time" is _actually the truth_. Since the parser _is_ producing a typeless token stream, I think it's appropriate that we reify accordingly. We probably shouldn't try to "fix" this problem in the wrong place, instead we should try to change what is true in the parser. (Though this is probably also hard, and there's that resolution that says we can't).

I can't see any good way to solve everything in this case. :(

One related thought is that we could allow any object-value through Typed OM (regardless of registration status), and in general allow custom properties to hold any value (not just token streams).

```
// (Don't know or care if --x is registered or not)
elm.attributeStyleMap.set('--x', CSS.px(10));
elm.attributeStyleMap.get('--x'); // Returns CSSUnitValue
```
In other words, the underlying value created from `CSS.px(10)` retains its type such that we can reify the same type back. If `--x` turns out to be registered with a syntax that doesn't match, it's invalid at computed value time. If it's not registered, it substitutes as the tokenized serialization of the value. Note that this would apply to css-variables, not css-properties-values-api. At least this gives us some minimal typeness for specified styles.

andruud:
@tabatkins We will probably need to think outside the box we've created for ourselves to proceed, so here goes:

We could make `@property` a thing which takes effect immediately during parsing:

```
@property --x {
  syntax: "<length>";
  inherits: false;
  initial-value: 0px;
}

--x: 10px; /* Actually parsed as <length> parse-time */
```

Let's invent a `[[declaredPropertySet]]` slot on the associated Document. This slot is like `[[registeredPropertySet]]`, except `@property` properties go in `[[declaredPropertySet]]` and `CSS.registerProperty` properties go in `[[registeredPropertySet]]`.

 1. We _may_ query `[[registeredPropertySet]]` during `setProperty`, specified value reification in Typed OM (and so forth), but _not_ the `[[declaredPropertySet]]` slot. The `[[declaredPropertySet]]` slot may be queried (if necessary) by computed-value APIs *ONLY*.
    - This means we _can_ continue to reify specified values according to what's registered with `CSS.registerProperty`. I don't think we _actually should_, as we'd be misrepresenting the truth in the CSS declaration block.
 2. `@property` rules take effect immediately in the parser, which means that the parser can actually parse subsequent declarations according to the syntax right away.
 3. Since we parse real values right away, they exist natively in the parsed representation, and hence specified values reify with proper types automatically.
 4. `@supports` can now detect whether a given `@property` rule worked. (Yay, fallback).
 5. We can reject declarations parse-time.
 6. Properties declared with `@property` still substitute as their computed values. This means that previously parsed custom properties must be reinterpreted computed-value time, if a property is re-registered:

```
@property --x {
  syntax: "<length>";
  inherits: false;
  initial-value: 0px;
}

--x: 10px; /* <length> */

@property --x {
  syntax: "<color>";
  inherits: false;
  initial-value: red;
}

/* --x is still a <length>, but is now invalid at computed-value time */
```

 7. If a property is _also_ registered with `CSS.registerProperty`, that registration wins, and any value parsed as a result of `@property` rules are reinterpreted computed-value time.

Open questions:

 8. How to determine whether a property inherits if multiple `@property` rules exist and disagree.
 9. Which types do we accept (via `.set()` etc) in TypedOM for `@property`-registered properties? We can't look at the at the `[[declaredPropertySet]]` slot. We could accept anything (see previous comment). Or we could invent something new, like associating the `[[declaredPropertySet]]` with the `StylePropertyMap`/'CSS declaration block' instead of the Document. That way, the information we need is always locally available whenever something is parsed.

emilio:
> We could make @property a thing which takes effect immediately during parsing:

That sounds not great... What happens if we `removeRule` the `@property` rule? Do we need to go find all the following declarations for that property and re-parse them as unparsed values? From which text, if we've parsed them as `<length>` previously?

andruud:
> That sounds not great...

[OK](https://memegenerator.net/img/images/15851983.jpg).

> What happens if we removeRule the @property rule? Do we need to go find all the following declarations for that property and re-parse them as unparsed values?

Yes, no, maybe. I don't know yet.

> From which text, if we've parsed them as <length> previously?

Note that custom properties must [already preserve their exact input for serialization](https://logs.csswg.org/irc.w3.org/css/2016-09-19/#e722375). Naturally this isn't spec'd anywhere. (I'm not implying that this automatically makes this `@property` behavior work).

emilio:
Sorry, to be clear I don't try to be harsh, just trying to point out stuff that probably makes something hard. I very much appreciate the work y'all do ^.^

andruud:
> ^.^

😹 

I had some ideas for how to deal with `deleteRule` which could avoid reparsing, but it doesn't matter, since `insertRule` is even worse. I guess we'd have to reparse. Meh.

andruud:
Or perhaps not allow `insertRule` / `deleteRule` operations such that it would affect values that are already parsed. :P

emilio:
So my issue with reparsing is that it's not only reparsing already-parsed stuff, but also potentially bringing back to life declarations that were overridden.

I think "operations such that it would affect values that are already parsed" is also pretty hard to detect for the same reason.

So for example, for the case where you have a length property called `--foo`, I assume that, if it's registered before that:

```css
div {
  --foo: 10px; // This one parses.
  --foo: red; // Gets dropped at parse-time, I guess?
}
```

If then I deleteRule the `@property` rule, I need to brin the `--foo: red` back from nowhere. Unless I'm misunderstanding the proposal.

andruud:
> potentially bringing back to life declarations that were overridden [...] from nowhere

Yeah, I know ... we can't do that.

> I think "operations such that it would affect values that are already parsed" is also pretty hard to detect for the same reason.

It could be more defensive, like "insertions/deletions of `@property` rules affecting a custom property that has been used anywhere, once, ever". This specifically might not be what we want ... but I think if we want these properties to behave natively ( would be nice, right? :-) ), we'd need to define the properties outside the "regular author scope" of CSS, or limit how/when `@property` rules appear (which can have almost the same effect).

tabatkins:
I'm ready to merge this; any additional concerns, Anders or Emilio?

The end result is just that we can't reify according to a grammar until computed-value time; any specified values show as CSSUnparsedValue like an unregistered property, or `"*"`-grammared registered property. And setting a property pays no attention to grammar in any API.

andruud:
👍 / 😢 (Mixed feelings, but this is probably for the better).

> any additional concerns

Just that it annoys me that the path forward to mitigating the TypedOM situation is so unclear right now. Unless you have some new insight to share, Tab?

> any API

Don't forget to also undo the change for two-param `CSS.supports`!

tabatkins:
Hey, so long as computed values still reify well, I'm happy enough. The losses are worth gaining `@property`.

tabatkins:
> Don't forget to also undo the change for two-param CSS.supports!

Done.

emilio:
> Hey, so long as computed values still reify well, I'm happy enough. The losses are worth gaining @property.

Yeah, I agree with this sentiment. Thanks for bearing with me y'all :)

PR #923, opened 2019-07-25
Improve wording on web-animation integrations

majido:
I still think we need more work in this area but 

- Update web-animations spec link to be https://drafts.csswg.org which is more recent
- Improve wording on how worklet animation interacts with web-animations
- Add section to deal with addition of "replace state" in Web animations.
- lots of small fixes
- Add issue to indicate our current shortcoming for async operations.


majido:
@flackr I plan to land multiple changes like this to the spec.

We can either do one-by-one reviews like this. Or I can ask you do a complete review at the end. I am inclined to think the latter would be more efficient and give us a better overall outcome. wdyt?

PR #935, opened 2019-08-20
Add explicit [Exposed]

saschanaz:
See also https://github.com/heycam/webidl/pull/423. (Found from https://github.com/web-platform-tests/wpt/pull/18382)


saschanaz:

PR #958, opened 2019-09-25
[worklets] Reject with script's error to rethrow

nhiroki:
This fixes #509

PR #967, opened 2019-10-02
[worklets] Queue a task to reject addModule()

### LINKED ISSUE #509: [worklets] addModule() should reject a promise with a more specific exception
The current spec defines that ```addModule()``` rejects a promise with "AbortError" DOMException in any case:

> 2. Reject promise with an "AbortError" DOMException.
> (see https://drafts.css-houdini.org/worklets/#fetch-and-invoke-a-worklet-script)

This obfuscates the reason why ```addModule()``` failed and makes it hard to debug. It would be nice to reject it with a more specific exception like "NetworkError" and "SyntaxError".

nhiroki:
This is a follow-up patch for #958. A task to reject addModule() should
be queued with script's error to rethrow to run on the thread where
addModule() is called.

This fixes #509

### DIFF
--- a/worklets/Overview.bs
+++ b/worklets/Overview.bs
@@ -449,14 +449,16 @@ When the user agent is to <dfn>fetch and invoke a worklet script</dfn> given |wo
 
             2. Reject |promise| with an "{{AbortError}}" {{DOMException}}.
 
-    4. If |script|'s <a>error to rethrow</a> is not null, then run these steps:
+    4. If |script|'s <a>error to rethrow</a> is not null, then <a>queue a task</a> on
+        |outsideSettings|'s <a>responsible event loop</a> given |script|'s <a>error to rethrow</a>
+        to run these steps:
 
         1. If |pendingTaskStruct|'s <a for="pending tasks struct">counter</a> is not <b>-1</b>, then
             run these steps:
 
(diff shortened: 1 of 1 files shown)


nhiroki:
@hiroshige-g @bfgeek Hi, can you review this?

worklets/Overview.bs
```diff
@@ -449,14 +449,16 @@ When the user agent is to <dfn>fetch and invoke a worklet script</dfn> given |wo
...
+    4. If |script|'s <a>error to rethrow</a> is not null, then <a>queue a task</a> on
+        |outsideSettings|'s <a>responsible event loop</a> given |script|'s <a>error to rethrow</a>
+        to run these steps:
 
         1. If |pendingTaskStruct|'s <a for="pending tasks struct">counter</a> is not <b>-1</b>, then
             run these steps:
 
             1. Set |pendingTaskStruct|'s <a for="pending tasks struct">counter</a> to <b>-1</b>.
 
-            2. Reject |promise| with |script|'s <a>error to rethrow</a>.
```
hiroshige-g: I assume the intention of this change is to reference the local argument passed to the task (i.e. avoid dereferencing |script| here), but `<a>error to rethrow</a>` still references `#concept-script`'s `error to rethrow`.

PR #974, opened 2019-12-10
Drop CSSPositionValue from README.md

sideshowbarker:
CSSPositionValue was removed from the spec in https://github.com/w3c/css-houdini-drafts/commit/5261c1a.


sideshowbarker:

PR #981, opened 2020-03-27
[worklets] Update agent and realm allocation

domenic:
Follows https://github.com/whatwg/html/pull/5411.

This moves to the imperative style of agent allocation, as HTML is no longer declaratively defining worklet agents and saying that they must be implicitly created at the appropriate time.


domenic:
This is ready to merge; the HTML changes have made their way into Shepherd.

PR #984, opened 2020-04-24
Handle module scripts that include top-level await.

Ms2ger:


Ms2ger:
This is in response to https://github.com/whatwg/html/pull/4352. @domenic , I'm not sure if this is what we want in terms of behavior, but it's a change, at least :)

worklets/Overview.bs
```diff
@@ -249,14 +251,19 @@ When a user agent is to <dfn export>create a WorkletGlobalScope</dfn>, given |wo
...
 
         Note: <a>Fetch a worklet script</a> won't actually perform a network request as it will hit
             the worklet's <a>module responses map</a>. It also won't have a parsing error as at this
             point it should have successfully been parsed by another worklet global scope. I.e.
             |script| should never be null here.
 
-    6. Run the <a>responsible event loop</a> specified by |insideSettings|.
+    9. [=Wait for all=] |scripts|, with success steps and failure steps both given by:
+
+        1. Run the <a>responsible event loop</a> specified by |insideSettings|.
```
domenic, line 266: I think this isn't right. Without an inner event loop, the microtask queue (and thus any awaits) cannot run.

I see two possible semantics here:

- Ignore the promises returned; their errors will be reported appropriately, and that's all one would do with them.
- Run the module scripts "sequentially", waiting for one to complete (i.e. the promise to settle) before moving on to the next.

I think the first of these is probably more correct, since (IIRC) it aligns with what happens if you have multiple TLA-using `<script type=module>`s.

worklets/Overview.bs
```diff
@@ -457,18 +464,21 @@ When the user agent is to <dfn>fetch and invoke a worklet script</dfn> given |wo
...
 
-            1. Decrement |pendingTaskStruct|'s <a for="pending tasks struct">counter</a> by
-                <b>1</b>.
+                    1. Decrement |pendingTaskStruct|'s <a for="pending tasks struct">counter</a> by
+                        <b>1</b>.
 
-            2. If |pendingTaskStruct|'s <a for="pending tasks struct">counter</a> is <b>0</b>, then
-                resolve |promise|.
+                    2. If |pendingTaskStruct|'s <a for="pending tasks struct">counter</a> is <b>0</b>, then
+                        resolve |promise|.
```
domenic, line 481: It looks like you shadowed an existing |promise| variable.

Ms2ger:
@bfgeek what do you think?

annevk:
This is now https://github.com/whatwg/html/pull/4352.

Issue #506, opened 2017-11-06
[worklets] dynamic import()


annevk:
I thought we already had an issue for this somewhere. If worklets cannot have `fetch()`, I don't think it makes sense for them to have dynamic `import()`. That would cause the same kind of execution problems.

cc @domenic @nhiroki @bfgeek

annevk:
See https://github.com/whatwg/fetch/pull/527#issuecomment-342038491 for issues if we decide to allow it somehow.

bfgeek:
I agree that it makes sense we shouldn't have `import()`.

bfgeek:
I'm not sure what the behaviour should be, if `import()` is just hidden, or it throws. :/

surma:
If we are disallowing it, I’d say hide it. A function whose only purpose is to throw seems odd.

domenic:
It's not a function; it's syntax. It can't be removed without language changes. I'm not even sure what hiding it would mean.

bfgeek:
Oh didn't realize that, makes sense.

annevk:
@domenic I assume we can make it reject the promise it returns? (Depending on what data gets passed to Fetch, we could even return a network error there for all fetches made by worklets. Not sure that's ideal however as it's unlikely to be implemented that way.)

domenic:
Yeah, that's doable.

nhiroki:
Filed a [tracking issue](https://bugs.chromium.org/p/chromium/issues/detail?id=782538) in crbug.

domenic:
To be clear, I don't think we _should_ reject import() in worklets, per my position from the earlier thread on the same topic. I think all forms of importing code should work in worklets, including lazily importing. I was only commenting on the technical feasibility here.

annevk:
I wonder how worklets work with the opaque origin setup if their modules can have non-dynamic imports. Did you review that @domenic? Or maybe the opaque origin is assigned after all dependencies are fetched?

The main reason to block dynamic `import()` is the same reason we block `fetch()`; to make it clear worklets are not supposed to have that kind of code.

nhiroki:
> I wonder how worklets work with the opaque origin setup if their modules can have non-dynamic imports. Did you review that @domenic? Or maybe the opaque origin is assigned after all dependencies are fetched?

[...]

PR #992, opened 2020-06-01
Define COEP for worklet environment settings object

yutakahirano:
This is part of https://github.com/whatwg/html/pull/5454.


yutakahirano:
The value doesn't matter IIUC as worklets don't make no-cors requests.

cc: @nhiroki @domenic @annevk

annevk:
As long as #506 gets fixed I think that's true.

yutakahirano:
@tabatkins Can you take a look?

PR #1016, opened 2020-12-16
[css-typed-om] Add missing parentheses for union

saschanaz:
This currently causes IDL parser failure.


foolip:
@tabatkins can you review and merge this?

PR #1017, opened 2020-12-16
[css-typed-om] Add missing semicolon after CSSColor constructor

foolip:


saschanaz:
Turns out there are two syntax issues at this line: #1016

foolip:
Impressive :)

PR #1020, opened 2021-02-16
[css-properties-values-api-1] Export registered custom property, etc

andruud:
- Export "registered custom property".
 - Export "universal syntax definition".
 - Fix a "the the".
 - When parsing against the syntax, we probably want "invalid at
   computed-value time", not guaranteed-invalid?

Somewhat related to: https://github.com/w3c/csswg-drafts/issues/5370


andruud:
cc @tabatkins

PR #1065, opened 2022-01-27
Add type checking to `CSSNumericValue.parse`

### LINKED ISSUE #1064: [css-typed-om] CSSNumericValue.parse() needs type checking for `<dimension-token>`
According to the [spec](https://drafts.css-houdini.org/css-typed-om-1/#dom-cssnumericvalue-parse), as long as the given string can be parsed into a `<dimension-token>`, the return value is the reification of the token. And [reification](https://drafts.css-houdini.org/css-typed-om-1/#reify-a-numeric-value) doesn't have type checking, either.

So `CSSNumericValue.parse('1xyz')` should return something non-null, which doesn't make sense, and is inconsistent with the [CSSUnitValue(value, type) constructor](https://drafts.css-houdini.org/css-typed-om-1/#dom-cssunitvalue-cssunitvalue), which throws when the type is invalid.

I think we should throw a `SyntaxError` when the unit is invalid. I'll put up a PR to fix it.

xiaochengh:
Fixes #1064

### DIFF
--- a/css-typed-om/Overview.bs
+++ b/css-typed-om/Overview.bs
@@ -1695,8 +1695,13 @@ rather than on {{CSSNumericValue}} instances.
         or a [=math function=],
         [=throw=] a {{SyntaxError}}
         and abort this algorithm.
+    
+    3. If |result| is a <<dimension-token>>
+       and [=creating a type=] from |result|'s unit returns failure,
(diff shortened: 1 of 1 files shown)


xiaochengh:
@tabatkins PTAL

PR #1068, opened 2022-04-10
[CSS-Typed-OM] Change reifying as from 'a CSSStyleValue' to 'an identifier'

DevSDK:
Step 2 of  parse()  [1] says "Reify a color value from  result ".

However, "Reify a color value" algorithm [2] needs a second parameter,

which is a property to be consumed by step 9 only.

Therefore, Change step 9 from 'CSSStyleValue' to 'an identifier'[3] for reifying.

And the reify a color value algorithm no more need |prop|.

Hence, Remove the |prop| parameter of 'reifying a color value'.

[1] https://drafts.css-houdini.org/css-typed-om-1/#dom-csscolorvalue-parse
[2] https://drafts.css-houdini.org/css-typed-om-1/#reify-a-color-value
[3] https://drafts.css-houdini.org/css-typed-om-1/#reify-an-identifier

Working Chromium CL: https://chromium-review.googlesource.com/c/chromium/src/+/3522201
csswg issue: https://github.com/w3c/csswg-drafts/issues/7174
cc. @xiaochengh

### DIFF
--- a/css-typed-om/Overview.bs
+++ b/css-typed-om/Overview.bs
@@ -5474,7 +5474,7 @@ CSS <<color>> values become either {{CSSColorValue}}s
 or generic {{CSSStyleValue}}s (otherwise).
 
 <div algorithm>
-    To <dfn>reify a color value</dfn> |val| for a property |prop|:
+    To <dfn>reify a color value</dfn> |val|:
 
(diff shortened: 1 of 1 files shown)


xiaochengh:
@tabatkins Could you ptal?

Issue #939, opened 2019-08-23
[css-properties-values-api] Should property registration be scoped?


smfr:
https://drafts.css-houdini.org/css-properties-values-api-1/

The fact that property registration is document global makes it impossible to use a custom property privately in a shadow tree. Maybe registration should be scoped somehow, optimally to a style sheet or shadow scope.

tabatkins:
I strongly agree that it would be great to scope these, in the same way as we intend to scope other CSS values. (And on that point, will someone *please* [either commit to this proposal or offer a credible alternative](https://github.com/w3c/csswg-drafts/issues/1995)?)

This means scoping to tree scopes, and the registration you use depends on the scope the stylesheet is in. (Tho this has some issues, which I'll discuss in a sec.)

> What happens when the property inherits from one scope into another, and the registration details change between scopes?

I think inheritance should happen with the synthesized token stream, as if you'd just set that token stream directly in the new context. Syntax has only a relatively small effect on registered properties, just making it go iacvt if necessary, and affecting how it reifies in Typed OM.

> What does it mean to use constructable style sheets, where the one CSSStyleSheet object could be inserted into multiple shadow trees, and the registration differs between shadow trees?

You already have to track the stylesheet separately in each location it's inserted into, as the separate locations can already clash and have to be resolved based on document order/shadow origin/etc. (Eg a stylesheet can be inserted into both the outer page and a shadow, targeting the same element with a `.foo` rule from the light side and a `:host` rule from the shadow side.)

------

I think the *biggest* conflict here is: what registration does an element use when it's a shadow host? That is, if the main page registers `--foo` with `syntax:"<length>"`, and the shadow registers it with `syntax:"<color>"`, if you set "--foo: red;" on a shadow host is it iacvt or a valid red?

Perhaps it matters based on where that rule comes from? Aka, a custom property's value carries the value's scope with it, and once you run the cascade and figure out which value wins, you know how to process it as a computed value? I think this matches with the proposed behavior in <https://github.com/w3c/csswg-drafts/issues/1995> for other properties -- setting `font-family: foo` can refer to either the light or shadow `@font { font-family: foo; }` based on whether the declaration is from the light or shadow tree.

However, it does mean that a script trying to use the Typed OM to interact with the property might be confused by what it gets! In my example, if the outer page asks for the value, it'll be expecting a CSSNumericValue, but it might instead get a CSSColorValue if the declaration from the shadow wins. Maybe that's okay?

[...]

PR #1085, opened 2023-01-31
[css-properties-values-api] Allow @property in shadow trees

andruud:


andruud:
@tabatkins: A question in the form of a PR: did you intend to lift this restriction in https://github.com/w3c/css-houdini-drafts/commit/ed18327ec9e7b898257c39db1ceb87a025110812 ?

It seems odd to have good rationale for a certain ShadowDOM behavior (2.8), and then _not_ allow `@property` in shadow trees.

tabatkins:
...yes, I *did* intend to lift the restriction in that other commit. Thanks for the catch. ^_^

foolip:
Looks like https://github.com/web-platform-tests/wpt/blob/master/css/css-properties-values-api/at-property-shadow.html will need to updated to match what the spec says.

emilio:
@andruud @tabatkins @astearns: Was dis discussed in the working group / was there a resolution on this?

It seems weird that we "leak" property definitions from shadow roots into the global scope. But also the ordering of the registrations etc needs to be defined. I don't think right now we have a dependency between the order of stylesheets inside vs. outside a shadow tree.

andruud:
> ordering of the registrations etc needs to be defined

Indeed, I've also wondered about this.

astearns:
@emilio I think this is a relevant resolution (but a bit vague) https://github.com/w3c/csswg-drafts/issues/1995#issuecomment-848941922

emilio:
@astearns there's no mention of `@property` there, nor in https://drafts.csswg.org/css-scoping-1/#shadow-names which was the section that discussion was about?

astearns:
@emilio more discussion of `@property`: https://github.com/w3c/css-houdini-drafts/issues/939

PR #1111, opened 2023-10-18
[css-properties-values-api] Use the appropriate ref link for "parse"

cdoublev:
I do not think the value of `initial-value` should be parsed with *consume a syntax definition*, but with *parse something according to a CSS grammar (aka simply parse)*.


cdoublev:
Fixed in [e1f11c5](https://github.com/w3c/css-houdini-drafts/commit/e1f11c5e6fffef319bcdc0f0fe25034d1a6a4b3f#diff-dbc90e119f2b577f29cada01a94b64b9258a0456d4fb1e6eda99ce1ece570828L652-L665).

PR #1114, opened 2023-10-30
[css-typed-om] Simplify CSSMathValues when creating internal rep.

andruud:
[#9451](https://github.com/w3c/csswg-drafts/issues/9451)


javifernandez:
@tabatkins do you think we could merge this ?

tabatkins (approved):

PR #1119, opened 2023-12-29
 [css-typed-om] Renamed 'flex' base type to 'fraction'

SebastianZ:
The `<flex>` data type got generalized and moved to CSS Values 5 and renamed to `<fraction>` in https://github.com/w3c/csswg-drafts/pull/9516.

So this adjustment also needs to be done in CSS Typed OM.
(Note that the type was renamed to 'fraction' to better distinguish it from Flexbox, because of the unit being called `fr`, and due to it getting more use cases outside of flexible grid track widths.)

Sebastian

### DIFF
--- a/css-typed-om/Overview.bs
+++ b/css-typed-om/Overview.bs
@@ -1025,7 +1025,7 @@ are represented by subclasses of the {{CSSNumericValue}} interface.
         "time",
         "frequency",
         "resolution",
-        "flex",
+        "fraction",
         "percent",
     };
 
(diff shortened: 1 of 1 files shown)


FremyCompany:
Did this ship anywhere yet?

SebastianZ:
> Did this ship anywhere yet?

Hm, looks like Chromium and WebKit already support `flex`. So I guess it is too late to rename the base type to 'fraction'. So I reverted that change. Though I kept the reference to the renamed data type.

Sebastian

PR #1127, opened 2024-08-09
Editorial: Update use of WebIDL "invoke a callback function"

jeremyroman:
Rethrowing is no longer implicit (since reporting is frequently what's desired). Here handling the exception is intended, so this is a very small change.

Also included are the various changes required to make these specs build and fixes to some places nearby where a clearly incorrect definition was linked to.

Part of whatwg/webidl#1425.

### DIFF
--- a/css-layout-api/Overview.bs
+++ b/css-layout-api/Overview.bs
@@ -70,6 +70,8 @@ urlPrefix: https://www.w3.org/TR/CSS21/; type:dfn
 urlPrefix: https://html.spec.whatwg.org/#; type: dfn
     text: structuredserializeforstorage
     text: structureddeserialize
+urlPrefix: https://webidl.spec.whatwg.org/; type: dfn
+    url: #is-a-platform-object; text: is a platform object
 </pre>
 
 Introduction {#intro}
@@ -352,8 +354,9 @@ is called, the user agent <em>must</em> run the following steps:
     5. Let |inputPropertiesIterable| be the result of [=Get=](|layoutCtor|, "inputProperties").
 
     6. If |inputPropertiesIterable| is not undefined, then set |inputProperties| to the result of
-        [=converting=] |inputPropertiesIterable| to a <code>sequence&lt;DOMString></code>. If an
-        exception is [=thrown=], rethrow the exception and abort all these steps.
+        [=converted to an IDL value|converting=] |inputPropertiesIterable| to a
+        <code>sequence&lt;DOMString></code>. If an exception is [=thrown=], rethrow the exception
+        and abort all these steps.
 
     7. Filter |inputProperties| so that it only contains [=supported CSS properties=] and [=custom
         properties=].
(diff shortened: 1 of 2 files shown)


jeremyroman:
@bfgeek ptal?

annevk:
cc @tabatkins

css-layout-api/Overview.bs
```diff
@@ -70,6 +70,8 @@ urlPrefix: https://www.w3.org/TR/CSS21/; type:dfn
 urlPrefix: https://html.spec.whatwg.org/#; type: dfn
     text: structuredserializeforstorage
     text: structureddeserialize
+urlPrefix: https://webidl.spec.whatwg.org/; type: dfn
+    url: #is-a-platform-object; text: is a platform object
```
tabatkins, line 74: Delete these lines, WebIDL is in the linking database and that anchor is linkable normally.

(It's just not exported, so you'd need to specify the spec manually in link-defaults right now, tho I'm fixing that in <https://github.com/whatwg/webidl/pull/1427>)

tabatkins (approved):
r+ once the manual "anchors" entry is removed

css-layout-api/Overview.bs
```diff
@@ -1969,20 +1975,20 @@ following steps:
 
             1. Let |fragmentResultValue| be |value|.
 
-    16. If |fragmentResultValue| is a [=platform object=]:
+    16. If |fragmentResultValue| [=is a platform object=]:
```
domenic, line 1978: This is a very strange way of doing it. I would expect `If |fragmentResultValue| [=implements=] {{FragmentResult}}`.

What the current spec text is doing means that

```js
const possibleValue = new Blob();
possibleValue.inlineSize = 5;
```

cannot be used. Whereas every other web API that accepts a dictionary, using normal Web IDL semantics, would allow it.

The root problem here appears to be overuse of the `Function` IDL callback type instead of declaring a callback that returns a proper Web IDL type, in this case `(FragmentResult or FragmentResultOptions)`.

I don't know what the current implementation is, and so I don't know if we can fix this. But I don't think we should export the internal "is a platform object" for use by external specs because of misuse like this.

jeremyroman: I can undo this line of change (though I think this text is trying to invoke that algorithm, even if we'd like it not to), if we prefer that to acknowledging that it's relying on a non-exported definition.

What I would prefer not to do is block this PR on a deep dive of CSS Layout API's choice here.

tabatkins: Yeah, we should not be discussing API design in this PR. If there's an issue, it should be raised separately; this is just a syntax transform.

domenic: My preference would be to merge this PR as-is and open a new issue. We should not block on deleting https://github.com/w3c/css-houdini-drafts/pull/1127#discussion_r1728015837

tabatkins: Deleting the manual anchor was just a "proper use of Bikeshed" thing; whether or not the anchor is exported was irrelevant, and just means it needed a link-defaults line instead to allow it grab a non-exported definition. I've fixed it.

PR #1137, opened 2025-01-02
[css-properties-values-api] Allow dashed idents in syntax strings

andruud:
Currently, grammars like `syntax:"auto | --mything"` are not allowed, since we don't handle HYPHEN-MINUS during "consume a syntax component". This was likely an oversight when it was initially written.

Note: we might want to just remove the algorithms that parse syntax components from a string, and instead tokenize/parse using the new `<syntax>` production. If we do, that would *also* make dashed idents allowed unless we explicitly block it.


andruud:
cc @kizu

PR #1138, opened 2025-01-17
Make PaintRenderingContext implement CanvasFilters

Kaiido:
This looks like an oversight, both implementations already do include that, and the below text doesn't mention it in the list of features that aren't included.

[Tests already exist](https://github.com/web-platform-tests/wpt/blob/master/css/css-paint-api/paint2d-filter.https.html).

I hope this doesn't qualify as a substantive change.


Kaiido:
That's actually a duplicate of https://github.com/w3c/css-houdini-drafts/pull/1055 which I forgot about...

PR #1139, opened 2025-03-07
[css-properties-values-api] Make var() fallback type-agnostic

andruud:
Resolves https://github.com/w3c/csswg-drafts/issues/10455.

### DIFF
--- a/css-properties-values-api/Overview.bs
+++ b/css-properties-values-api/Overview.bs
@@ -400,16 +400,6 @@ and [[css-syntax-3#tokenization|tokenizing]] the resulting string.
 	a <<dimension-token>> with a value of "80" and unit "px".
 </div>
 
-### Fallbacks In ''var()'' References ### {#fallbacks-in-var-references}
-
-References to registered custom properties using the ''var()'' function may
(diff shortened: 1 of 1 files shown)


andruud:
I'd like to ship this while usage is still low ([Blink I2S](https://groups.google.com/a/chromium.org/g/blink-dev/c/yc4XmGZi1Mk)). cc @emilio and @anttijk (maybe?)

emilio (approved):
Do you know if tests for this have already been updated?

emilio:
We should probably move this spec to csswg-drafts proper at this point btw.

andruud:
> Do you know if tests for this have already been updated?

I checked this morning, and they were not updated then. (I will do it when I make the change in Blink.)

emilio:
Ok, filed https://bugzilla.mozilla.org/show_bug.cgi?id=1952516 for this. If you have the test changes around feel free to send them, otherwise I can fix them up next week or just wait for your change, as you prefer (I'm on PTO technically today, so about to go enjoy the sun a little bit :P)

PR #1163, opened 2026-01-22
[css-properties-values-api][editorial] CSS.registerProperty() requires inherits argument

cdoublev:
The [register a custom property algorithm](https://drafts.css-houdini.org/css-properties-values-api-1/#register-a-custom-property) is defined as taking `inherits` as an optional argument whereas it is passed from the [`PropertyDefinition`](https://drafts.css-houdini.org/css-properties-values-api-1/#dictdef-propertydefinition) dictionary, which defines it as `required`.

  > To register a custom property with `name` being a string, and optionally `syntax` being a string, `inherits` being a boolean, and `initialValue` being a string, ...

  > ```
  > dictionary PropertyDefinition {
  >   required DOMString name; 
  >   DOMString syntax       = "*"; 
  >   required boolean   inherits; 
  >   DOMString initialValue; 
  > };
  > ```

### DIFF
--- a/css-properties-values-api/Overview.bs
+++ b/css-properties-values-api/Overview.bs
@@ -694,9 +694,9 @@ as arguments of the same names.
 <div algorithm>
 	To <dfn>register a custom property</dfn>
 	with |name| being a string,
+	|inherits| being a boolean,
 	and optionally
 	|syntax| being a string,
-	|inherits| being a boolean,
 	and |initialValue| being a string,
 	execute these steps:
 


cdoublev:
Obsoleted by [e1f11c5](https://github.com/w3c/css-houdini-drafts/commit/e1f11c5e6fffef319bcdc0f0fe25034d1a6a4b3f#diff-dbc90e119f2b577f29cada01a94b64b9258a0456d4fb1e6eda99ce1ece570828L688) (`PropertyDefinition.inherits` is now optional).

PR #1177, opened 2026-07-30
CSS Factories exposed to Worker context

javifernandez:
The CSS namespace is only exposed to Window context. The CSS factories are very convenient to implement tests of CSS TypedOM features that are exposed to Workes.

This PR makes the CSS namespace extension proposed by CSS TypedOM to be exposed to Worker contexts.

It's been discussed in the [csswg issue 14229](https://github.com/w3c/csswg-drafts/issues/14229) in the context of allowing partial extensions of the CSS namespace to be exported to contexts different than Window. 

The WPT [PR 61330](https://github.com/web-platform-tests/wpt/pull/61330) modifies some of the current CSS TypedOM tests so that they are run in Workers, requiring these CSS Factories to be exported to Worker contexts as well.


css-typed-om/Overview.bs
```diff
@@ -2382,6 +2382,7 @@ to create new numeric values much less verbosely
 than using the constructors directly.
 
 <xmp class=idl>
+[Exposed=(Window, Worker)]
 partial namespace CSS {
```
emilio, line 2386: This should be `[Exposed=(Window, Worker, PaintWorklet, LayoutWorklet)]`, but also I think a partial with a wider exposure than the namespace itself is not legal in WebIDL so we should fix CSSOM and co too.

javifernandez: I have realized that this would require a cross-spec PR to change the exposure of the different namespace CSS partial definitions, so that we declared them Window only explicitly. In addition, you have risen the concern of future partial definitions not being ware of the Worker exposure by default. 

I'm already working on such PR, as it seems there was some agreement in [issue 14229](https://github.com/w3c/csswg-drafts/issues/14229), but perhaps this needs more discussion ? When we merged the [PR 61330)](https://github.com/web-platform-tests/wpt/pull/61330) for the tests I thought this spec change was easier.

javifernandez: I have submitted the [PR 14245](https://github.com/w3c/csswg-drafts/pull/14245) to fix the CSSOM and co, as requested. 

That PR for cssom-1 would make the explicit exposure of the css-typed-on-1 partial definition redundant; even more, if eventually the CSS namespace exposure in cssom-1 is shortened, the css-typed-om-1 declaration won't be compliant with the WebIDL subset-rule about partials.

emilio (approved):

